PR #62, opened 2023-03-31
🛑URGENT🛑 Capitalize Tweet for consistency in README.md

slweeb:
Not every instance of the word "Tweet" was capitalized.


lucasnethaj:
thank you for your service

issamoxix:
lgtm

alsonick:
LGTM

GalvinPython:
lgtm

hoangphuc05:
lgtm

melon-husk:
lgtm

brandondrew:
are you sure that the solution is to use upper case on all of them, rather than to use lower case on all of them?

slweeb:
> are you sure that the solution is to use upper case on all of them, rather than to use lower case on all of them?

Yes, official Twitter style is to capitalize "Tweet", such as in the Help Center: https://help.twitter.com/en/using-twitter/how-to-tweet

hoangphuc05:
I think we should lowercase all "Tweet" in the Help Center instead

slweeb:
> I think we should lowercase all "Tweet" in the Help Center instead

It's not just the Help Center, it's every Twitter article ever, it's the standard.

brandondrew:
> > are you sure that the solution is to use upper case on all of them, rather than to use lower case on all of them?
> 
> Yes, official Twitter style is to capitalize "Tweet", such as in the Help Center: https://help.twitter.com/en/using-twitter/how-to-tweet

In that case I support your PR

slweeb:
why was this closed? not fixed

hoangphuc05:
I told you, the correct direction is to lowercase “tweet” everywhere else. Closing this PR since the fix does not address the root cause

PR #235, opened 2023-03-31
Update Hit.java

Yeabsra-Henok:
The "metadata" field is now declared as a private
The "getMetadata" method now returns the "metadata" field directly The "setMetadata" method now takes a "ThriftSearchResultMetadata" parameter instead of a nullable version of the same parameter, The "isHasExplanation" method is renamed to "hasExplanation" for consistency with the naming conventions of boolean properties inJava :)


csubj:
your formatter makes it hard to tell what you changed - will probably be rejected for that reason

PR #242, opened 2023-03-31
Fix has_gte_10k_favs typo in CandidatePredicates

josegomezr:
Fixed CandidatePredicate string: said "has_gte_10k_favs" but the actual predicate checked for 1000.


vvFiCKvv:
nice catch!

DanyPell:
This is great finally a serious one : )

PR #282, opened 2023-03-31
Update abusive_model.py

bbriggs1990:
Simplify the loop for iterating over physical_devices


Smission:
lgtm

Dev-Tarek:
it's cleaner to store the list in a new variable;
you can append something to it or change the source without having to changing the loop definition

Smission:
@Dev-Tarek I think this is a preference more than anything and if done consistently, neither should be a problem. Also, even though insignificant, one might argue that declaring a variable only to use it once is wasting memory...

Dev-Tarek:
> @Dev-Tarek I think this is a preference more than anything and if done consistently, neither should be a problem. Also, even though insignificant, one might argue that declaring a variable only to use it once is wasting memory...

for sure it is a preference, but it's not worth the time to create a the PR or call it simplification imo.
for sure i understand that most of the PRs in the ongoing marathon are just for fun :D

bbriggs1990:
@Dev-Tarek @Smission It was just for fun lol. If only I knew removing this lone variable would be so controversial! XD

puskuruk:
Just sad

puskuruk:
It's just sad

PR #296, opened 2023-03-31
Feat: Improve build process and reduce code bloat by 33%

eahrend:
Make it exit 1 because 1 is larger (and therefore better than) 0

Also reducing lines of code by 33%


dcow:
Please exit 2, because 2 is the only even prime number.

eahrend:
@dcow 2 is twice as large as 1, and therefore takes up more memory. We're trying to optimize for performance here.

eahrend:
@dcow 2 is twice as large as 1, and therefore takes up more memory. We're trying to optimize for performance here.

Heliodex:
Duplicate of #23587012987199

matthewdeanmartin:
Build script is really gotten out of hand, can we reduce line code count?

eahrend:
> Build script is really gotten out of hand, can we reduce line code count?

Great idea. Will update

eahrend:
@matthewdeanmartin reduced line count by 33%

PR #304, opened 2023-03-31
Code linting: sort python imports using Ruff

harshil21:
The python code is poorly organized and unsorted, I just ran [ruff](https://github.com/charliermarsh/ruff) in the repository with `--extend-select I` option to sort the imports. I also excluded the twml directory since that is legacy code which is going to be removed anyway (right?)

The python code still needs lots of refactoring, if this is okay, I'm happy to set up ruff and [black](https://github.com/psf/black) in this repository.


harshil21:
@twitter-service can you give a reason on why this PR was closed? Not up to your standards? Or do you guys not accept any code style fixes and only algo improvements? I hope that isn't true.

dagelf:
No it's because their main branch was force pushed invalidating all PRs

harshil21:
Okay, I will try again, hope they communicate that better in the future.

PR #315, opened 2023-03-31
Add other OS in the game not just one

gekkowrld:
Only one OS file, it doesn't feel inclusive at all. So I added the others in, Anyway the list is definitely not complete!


MathisP75:
Oh lol, so the devs are not chad Linux users, but Mac users 🤮

PR #322, opened 2023-03-31
refactor: use more descriptive language

lukeify:
Hello.

Utilising a proper noun in this way is not ideal because at scale, there are many Elon's in this world. This PR updates the language to make sure we're only referring to the specific individual.


ivyturner (commented):
seems legit

abui-am:
upvote

PR #328, opened 2023-03-31
Changed the MaxTweetToFetch value from 500 to 420

dev3Nigerian:
Why 500 when it can be 420? Be cool...


anurag-roy:
In 19 days

smiba:
LGTM, ship it 🔥🔥🔥

PR #345, opened 2023-03-31
Remove censorship outside of adult content

randiaz95:
Censorship, the act of suppressing or controlling information, is detrimental to a healthy and democratic society. The freedom of expression and access to information are fundamental human rights that should be protected and encouraged. When censorship is imposed, it stifles diverse viewpoints, hinders creativity, and prevents the dissemination of important information that can lead to positive change. It can also lead to a lack of accountability, as those in power may hide information to avoid scrutiny or criticism. Ultimately, censorship undermines the democratic principles of transparency and accountability, limits our ability to learn and grow, and threatens the foundation of a free and open society.

This change only allows adult content to be considered offensive; thereby removing big brother.

### DIFF
--- a/src/java/com/twitter/search/common/relevance/classifiers/TweetOffensiveEvaluator.java
+++ b/src/java/com/twitter/search/common/relevance/classifiers/TweetOffensiveEvaluator.java
@@ -43,13 +43,11 @@ import com.twitter.search.common.util.text.TokenizerHelper;
 public class TweetOffensiveEvaluator extends TweetEvaluator {
   private static final Logger LOG = LoggerFactory.getLogger(TweetOffensiveEvaluator.class);
 
-  private static final int MAX_OFFENSIVE_TERMS = 2;
+  private static final int MAX_OFFENSIVE_TERMS = 3;
 
   private final File filterDirectory;
   private static final File DEFAULT_FILTER_DIR = new File("");
   private static final String ADULT_TOKEN_FILE_NAME = "adult_tokens.txt";
-  private static final String OFFENSIVE_TOPIC_FILE_NAME = "offensive_topics.txt";
-  private static final String OFFENSIVE_SUBSTRING_FILE_NAME = "offensive_substrings.txt";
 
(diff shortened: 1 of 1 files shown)


datkat21 (commented):
LGTM

anurag-roy:
d8f66d222a6e7a670c02af9d89a2d70e2c8881a101f5ab347a695649382bf352_1

randiaz95:
Consider this TLDR: This change only allows adult content to be considered offensive; thereby removing big brother.

anurag-roy:
Thanks for the tldr @randiaz95, makes sense now! LGTM!

paulbauriegel:
Filtering offensive topics is not a bad thing and has nothing to do with censorship from my point of view. The problem is that the text files are not open sourced, so anything could be written in them, but from that perspective you could also put your "censorship terms" into the adult_tokens.txt ...

randiaz95:
Yes, and that would be explicit censorship by putting non adult_tokens into the adult_tokens file.

Whereas now they can get away with it.

Lucretiel:
This change also causes even adult tokens to be ignored if they appear in usernames / display names

randiaz95:
If you look at the `isUserNameOffensive` method: Adult tokens were never being checked here for username, I can update to use the adult filter on the usernames instead though.

We should consider adult tokens being bad in usernames, if you search an adult token word in twitter you will get some crazy/raunchy stuff I likely don't think we will ever be able to filter.

randiaz95:
Closing this PR as I don't get paid to work for twitter.

Just use this as proof that we are being censored.

paulbauriegel:
I think filtering offensive content is not censorship, you are also beeping out all the f***s in american sitcoms

randiaz95:
> I think filtering offensive content is not censorship, you are also beeping out all the f***s in american sitcoms

Actually thats why I bumped the max offensive terms limit to 3, so we can still write "You F***ing *sshole"

Maybe if we can see a peepee going in a vajaijai with a simple twitter search we shouldn't censor things the FBI wants us to censor.

PR #359, opened 2023-03-31
❄️  Add Nix flake and pre-commit hooks

aaronmondal:
Setting up the different tools for development can be hard for new contributors. This commit adds a Nix flake that provides a reproducible development environment which is easy to set up and ensures consistent code style across all subprojects.

The initial flake contains `git`, `pre-commit`, `which` and `bazelisk` wrapped as a `bazel` executable. Dependencies like Rust, Scala etc may be added in the `packages` attribute in the future.

The current pre-commit configuration is a no-op. Future commits may enable the currently disabled parts of `pre-commit-hooks.nix` for a gradual transition.

Supports `(x86_64|aarch64) (Linux|MacOS)` systems.


aaronmondal:
This also allows rather straightforward integration with GitHub actions. See https://github.com/eomii/rules_ll/blob/main/.github/workflows/pre-commit.yml for an example that uses the same logic as implemented in this flake to run the hooks from a local GitHub action cache.

PickleNik:
Screenshot 2023-03-31 at 6 21 25 PM

PickleNik:
image

Issue #236, opened 2023-03-31
author_is_elon


enzo-js:
the problem is his tweets suck. stop recommending them. https://github.com/twitter/the-algorithm/blob/7f90d0ca342b928b479b512ec51ac2c3821f5922/home-mixer/server/src/main/scala/com/twitter/home_mixer/functional_component/decorator/HomeTweetTypePredicates.scala#L225

matteomad1011:
bro. too much

jcolicchio:
Concerning.

BradFitz66:
looking into it.

sohdas:
Interesting

nikolqy:
How much can I pay to get my tweets promoted like that?

kimjoaoun:
> How much can I pay to get my tweets promoted like that?

44 biliions, in cash.

nikolqy:
> > How much can I pay to get my tweets promoted like that?
> 
> 44 biliions, in cash.

I think you mean $20 billion now!

alex-redacted:
Cannot believe he got his own special tweet booster but he owns the thing, so I guess Melon can do whatever he wants.

fadyanwar:
Github Issues have never been so entertaining, I guess I will ditch reddit in its favour

diegololia:
Seems like a security vulnerability

rifusaki:
> Github Issues have never been so entertaining, I guess I will ditch reddit in its favour

Let's ditch Twitter for GitHub Issues too

Galluno:
real

nonbinarypotus:
based

fadyanwar:
> > Github Issues have never been so entertaining, I guess I will ditch reddit in its favour
> 
> Let's ditch Twitter for GitHub Issues too

Or maybe fork twitter and create a better open platform, hopefully Elon did see this coming...

couldbejake:
This only represents a small part of Twitter, and as others have pointed out, code has been stripped out.

PacerBridge:
Again, isn't this code purely just for logging stats?

vogelarthur:
This specific snippet is only for metrics...

https://github.com/twitter/the-algorithm/blob/7f90d0ca342b928b479b512ec51ac2c3821f5922/home-mixer/server/src/main/scala/com/twitter/home_mixer/functional_component/feature_hydrator/RequestQueryFeatureHydrator.scala#L87

gsuhm:
!!

[...]

PR #389, opened 2023-03-31
Fixes issue #236 pertaining to a priority bug.

LarissaMayrink:
Line 192 call led to a branch which contained a unwarranted definition.

_I have no idea how that ended up in there_ -54 people that already reported it


uwussimo (commented):
This is the best pull request possible!!!!!

PR #437, opened 2023-03-31
Exclude Auto-generated .vscode Folder by Default

soharaa:
This pull request enhances the usability of the development workflow by providing a convenient solution for developers who use Visual Studio Code (VS Code) as their main development environment. With this update, users can now use VS Code without having to worry about accidentally committing their potentially auto-generated .vscode folder. This improvement simplifies the process of sharing and collaborating on code, while also promoting best practices for version control. Overall, this pull request streamlines the development experience for VS Code users, enabling them to focus on what matters most: writing high-quality code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .DS_Store
 
+.vscode


martinwoodward:
Rather than editor specific files in the repo, we would generally encourage users of a particular editor (such as VS Code) to add temporary folders like that to their machine wide global git config.  See the following for more information:

https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

Of course, if it becomes a problem for the maintainer that these temp files are getting regularly submitted in PR's then they might want to consider adding it anyway.  But ideally individuals would configure that on a per machine basis especially when a specific editor is not being mandated by the project to prevent the gitignore containing every editors cruft files.

gekkowrld:
What about OS specific files, are they the same, because it seems like MacOs has some special privileges

busybox11:
> What about OS specific files, are they the same, because it seems like MacOs has some special privileges 

`.DS_Store` is a file that is automatically created in EVERY SINGLE directory that has ever been opened by Finder at some point.
Its purpose is to remember the position of the icons representing its content, which is not really useful in a dev project.
It's usually ignored with git to prevent bloat inside the repo, since this file basically serves no purpose at all for us (except if actually specified).

AnEmortalKid:
Have you tried using vim instead ?

PR #439, opened 2023-03-31
prefer folds over pattern matching

Ghurtchu:
pattern matching boilerplate:
```scala
val strResult = Option(5) match { 
  case Some(v) => v.toString
  case None => "0"
}
```

fold:
```scala
val strResult = Option(5).fold("0")(_.toString)
```

### DIFF
--- a/ann/src/main/scala/com/twitter/ann/hnsw/JMapBasedIdEmbeddingMap.scala
+++ b/ann/src/main/scala/com/twitter/ann/hnsw/JMapBasedIdEmbeddingMap.scala
@@ -44,10 +44,7 @@ private[hnsw] object JMapBasedIdEmbeddingMap {
     injection: Injection[T, Array[Byte]],
     numElements: Option[Int] = Option.empty
   ): IdEmbeddingMap[T] = {
-    val map = numElements match {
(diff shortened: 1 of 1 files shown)


ElijahPepe:
This is a style guide question for @mariusae, assuming he still works there. In my personal opinion, `fold()` is less readable than pattern matching. From my knowledge of Scala, pattern matching tends to be a little more efficient and easier to read. This repository also uses `getOrElse()` for a few variables, so perhaps that's another option (although I find `getOrElse()` to be semantically ambiguous; the type of some variables can be obscured).

Ghurtchu:
> This is a style guide question for @mariusae, assuming he still works there. In my personal opinion, `fold()` is less readable than pattern matching. From my knowledge of Scala, pattern matching tends to be a little more efficient and easier to read. This repository also uses `getOrElse()` for a few variables, so perhaps that's another option (although I find `getOrElse()` to be semantically ambiguous; the type of some variables can be obscured).

It is not harder to read it is just unfamiliar like any other thing people don't know. When I became familiar with `fold` I never went back to `pattern matching` or `map` followed by `getOrElse`.

Same can be said on `Either`, `Try` or any other `Foldable` data structure.

PR #472, opened 2023-03-31
remove unnecessary braces and indentation

Ghurtchu:
Hit dislike button if you like this PR.

### DIFF
--- a/visibilitylib/src/main/scala/com/twitter/visibility/rules/VisibilityPolicy.scala
+++ b/visibilitylib/src/main/scala/com/twitter/visibility/rules/VisibilityPolicy.scala
@@ -91,13 +91,7 @@ object VisibilityPolicy {
     BounceQuotedTweetTombstoneRule
   )
 
-  def union[T](rules: Seq[Rule]*): Seq[Rule] = {
-    if (rules.isEmpty) {
-      Seq.empty[Rule]
-    } else {
-      rules.reduce((a, b) => a ++ b.filterNot(a.contains))
-    }
-  }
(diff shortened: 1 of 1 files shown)


BenMusch:
LGTM, i recommend we change all variable names to 1-letter to save disk space as well

visibilitylib/src/main/scala/com/twitter/visibility/rules/VisibilityPolicy.scala
```diff
@@ -91,13 +91,7 @@ object VisibilityPolicy {
...
   )
 
-  def union[T](rules: Seq[Rule]*): Seq[Rule] = {
-    if (rules.isEmpty) {
-      Seq.empty[Rule]
-    } else {
-      rules.reduce((a, b) => a ++ b.filterNot(a.contains))
-    }
-  }
+  def union[T](rules: Seq[Rule]*): Seq[Rule] = if (rules.isEmpty) Seq.empty[Rule] else rules.reduce((a, b) => a ++ b.filterNot(a.contains))
```
shailrshah, line 94: IMO this line is too long. I'd stick to a max of 120 characters per line. 

> * If lines are too short, text becomes hard to read because you must constantly jump from one line to the next while reading.
> * If lines are too long, the line jumping becomes too hard because you "lose the line" while going back to the start of the next line (this can be mitigated by having a bigger inter-line spacing, but this also wastes space).

* https://softwareengineering.stackexchange.com/a/222998
* https://docs.scala-lang.org/style/indentation.html

PR #491, opened 2023-03-31
Remove busy wait in logging loop

xxDark:
Replace polling with standard `BlockingQueue#take()` method.


shailrshah (commented):
LGTM, nice use of `take()`. 

https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/BlockingQueue.html#take()

PR #495, opened 2023-03-31
proof-read comments

dev3Nigerian:
I proof-read the comments with emphasis on the interchangeability of the words "it's" and "its".

"It's" is a contraction and should be used where a sentence would normally read "it is".
"Its" with no apostrophe, on the other hand, is the possessive word.

I made some changes to these words where necessary in the comments. I also made some minor corrections and the occasional full-stop. 

Kind Regards.


Psychpsyo:
"it's" is also a contraction that can be used where a sentence would normally read "it has".
And the very first line,
> \- if candidate is discoverable given that it's from an address-book/phone-book based source
should stay as "it's" regardless since it does mean "it is" here.

follow-recommendations-service/common/src/main/scala/com/twitter/follow_recommendations/common/predicates/gizmoduck/GizmoduckPredicate.scala
```diff
@@ -39,8 +39,8 @@ import javax.inject.Singleton
 
 /**
  * In this filter, we want to check 4 categories of conditions:
- *   - if candidate is discoverable given that it's from an address-book/phone-book based source
- *   - if candidate is unsuitable based on it's safety sub-fields in gizmoduck
+ *   - if candidate is discoverable given that its from an address-book/phone-book based source
```
mapgie, line 42: "given that it is from an address- book"

"it's" is OK.

dev3Nigerian: You are right. An oversight on my part

PR #496, opened 2023-03-31
fix: add missing dots to `e.g.` and `i.e.`

PeasfulTown:
There are instances of `e.g` `eg.` `eg,` `eg` `ie.` `i.e` `ie.` `ie` within this repository.

This fix adds missing dots to `e.g.` and `i.e.` which is the grammatically correct form of these expressions.


Marioood:
LGTM

PR #497, opened 2023-03-31
Change Refactor

ImperiumTakp:
inverted if and changed for comprehension to a foldRight with a match on the result

### DIFF
--- a/src/scala/com/twitter/simclusters_v2/scalding/KnownForSources.scala
+++ b/src/scala/com/twitter/simclusters_v2/scalding/KnownForSources.scala
@@ -102,28 +102,30 @@ object KnownForSources {
     TypedPipe
       .from(TextLine(textFile))
       .flatMap { str =>
-        if (str.startsWith("#")) {none}
-        try {
-          val tokens = str.trim.split("\\s+")
-          val userId = tokens(0).toLong
-          (1 until tokens.length).foldRight(Array.newBuilder[(Int, Float)])((i, r) => {
-            val Array(cIdStr, scoreStr) = tokens(i).split(":")
-            val clusterId = cIdStr.toInt
-            val score = scoreStr.toFloat
-            val newEntry = (clusterId, score)
-            r += newEntry
-          }).result() match {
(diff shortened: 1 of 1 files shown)


src/scala/com/twitter/simclusters_v2/scalding/KnownForSources.scala
```diff
@@ -102,31 +102,29 @@ object KnownForSources {
...
-              Some((userId, res.result()))
-            } else None
-          } catch {
-            case ex: Throwable =>
-              log.warning(
-                s"Error while loading knownFor from $textFile for line <$str>: " +
-                  ex.getMessage
-              )
-              None
+        if (str.startsWith("#")) {none}
```
Ghurtchu, line 105: `none`? I think you need to fix that to `None`.

Ghurtchu: Additionally, you don't need curly braces, so it can be rewritten like:
```scala
if (str.startsWith("#")) None
```

ImperiumTakp: just redid it with a match instead of the if. Have you tested the code locally?

>

Ghurtchu: Yep, before you inverted the logic it was like:
```scala
else None
```
So it's also legit to do:
```scala
if (true) None
```.

Comment (ImperiumTakp) [reply src/scala/com/twitter/simclusters_v2/scalding/KnownForSources.scala]: yes the main logic should be in the bottom of the block, and short to no result

File: src/scala/com/twitter/simclusters_v2/scalding/KnownForSources.scala [hunk 2]
```diff
@@ -103,7 +103,7 @@ object KnownForSources {
       .from(TextLine(textFile))
       .flatMap { str =>
         str match {
-          case s"#$_" => none
+          case s"#$_" => None
```
Ghurtchu, line 106: Superb!

innoalex (commented):
Good job, LGTM

PR #514, opened 2023-04-01
fix grammar in simclusters-ann readme.md and ScoreFacadeStore.scala

animusDS:
The use of the word "Consider" in the sentence "Considering that we need to process over 6,000 RPS, it’s hard to support by the existing infrastructure. " is wrong.

The proper word is "Considering"


sdornan:
Should be “with the existing infrastructure” too.

PR #519, opened 2023-04-01
prefer fold over map followed by getOrElse

Ghurtchu:
`fold` combinator is a shorthand way of encoding `map` + `getOrElse` pattern.


AnEmortalKid:
While true, I think this reduces readability for those not familiar with these higher order functions. 

Remmeber, this is supposed to be the public square where everyone can contribute. If we’re adding a barrier of knowing what a fold / fold left is, it’s going to dissuade contributors that are not familiar with functional programming.

Ghurtchu:
> While true, I think this reduces readability for those not familiar with these higher order functions.
> 
> Remmeber, this is supposed to be the public square where everyone can contribute. If we’re adding a barrier of knowing what a fold / fold left is, it’s going to dissuade contributors that are not familiar with functional programming.

I don't think `fold` is complicated enough to discourage someone. This is a chance for them to learn a little about it, why not? `map` + `getOrElse` is also functional programming, just a little weird way of encoding `fold`.

PR #520, opened 2023-04-01
Align comments to same column

thescripted:
Keep comments on same column because it looks nice


SergeyYurievich (commented):
I reviewed the changes and i can approve that it looks nicer

emanuallan:
FISI

PR #528, opened 2023-04-01
refactor(EarlyBirdIndexConfig): implements strategy pattern to create con…

lmaths:
-  Refactor method "createEarlybirdIndexConfig" to use a strategy pattern to createconfigs.


Matheus-Silas97:
it's a great solution, we have try refactor if else methods using this pattern too

MarlonnCarvalhosa:
Great refactoring, we saw that the team did not take the matter very seriously.

PR #533, opened 2023-04-01
Make an optimization for BatchingClient

xxDark:
This optimization avoids allocation of new hash table on each batch being sent by using an array storing original content of requests to send. Not sure how often this code path is taken, but based off of the comment left originally, I decided to make this.
Not sure whether this is possible since I don't know which type `GetTweetsRequest#setTweet_ids` accepts as the parameter and how its fields are serialized (I assume its parameter is a plain `List`, not an `ArrayList` specifically).

### DIFF
--- a/src/java/com/twitter/search/ingester/pipeline/util/BatchingClient.java
+++ b/src/java/com/twitter/search/ingester/pipeline/util/BatchingClient.java
@@ -2,7 +2,7 @@ package com.twitter.search.ingester.pipeline.util;
 
 import java.util.HashSet;
 import java.util.Map;
-import java.util.Set;
+import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 
 import com.google.common.collect.Sets;
@@ -20,7 +20,7 @@ public class BatchingClient<RQ, RP> {
     /**
      * Issue a request to the underlying store which supports batches of requests.
      */
-    Future<Map<RQ, RP>> batchGet(Set<RQ> requests);
+    Future<Map<RQ, RP>> batchGet(List<RQ> requests);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


abetwothree:
Does this PR help Elon pay off 44 billion? If so, merge. If not, merge anyways and let's see what happens.

ludo6577:
Maybe they need it ordered: https://stackoverflow.com/questions/1035008/what-is-the-difference-between-set-and-list

xxDark:
> They were using a hashset in the first place which is not ordered by definition

src/java/com/twitter/search/ingester/pipeline/util/ResponseNotReturnedException.java
```diff
@@ -2,6 +2,6 @@
 
 public class ResponseNotReturnedException extends Exception {
   ResponseNotReturnedException(Object request) {
-    super("Response not returned in batch for request: " + request);
+    super("Response not returned in batch for request: " + request, null, false, false);
```
Taz03, line 5: explain this, seems unnecessary to me

src/java/com/twitter/search/ingester/pipeline/util/BatchingClient.java
```diff
@@ -63,27 +63,25 @@ private void maybeBatchCall(RQ request) {
 
       // Make a copy of requests so we can modify it inside executeBatchCall without additional
       // synchronization.
-      frozenRequests = new HashSet<>(unsentRequests);
+      frozenRequests = Arrays.asList((RQ[]) unsentRequests.toArray());
```
Taz03, line 66:
```suggestion
      frozenRequests = List.of(unsentRequests.toArray(new RQ[0]));
```
this better, pls commit the suggestion i want to upgrade my pea badge

xxDark: Does not compile.

src/java/com/twitter/search/ingester/pipeline/util/ResponseNotReturnedException.java
xxDark: Disables stack trace generation.

PR #543, opened 2023-04-01
simplify Option expression

Ghurtchu:
`collectFirst` is a deprecated method, I believe my suggestion is better.


PrestigePvP:
LGTM

PR #568, opened 2023-04-01
fix code smells

laoal11:
fixed a few code smells and trying to make the code a bit more readable at times

### DIFF
--- a/src/java/com/twitter/search/core/earlybird/facets/AbstractFacetCountingArray.java
+++ b/src/java/com/twitter/search/core/earlybird/facets/AbstractFacetCountingArray.java
@@ -93,8 +93,7 @@ public abstract class AbstractFacetCountingArray implements Flushable {
 
   protected static final int decodeTermID(int facetID) {
     if (facetID != UNASSIGNED) {
-      int termID = facetID & TERM_ID_MASK;
-      return termID;
+      return facetID & TERM_ID_MASK;
     }
 
     return EarlybirdIndexSegmentAtomicReader.TERM_NOT_FOUND;
--- a/src/java/com/twitter/search/earlybird/archive/ArchiveSearchPartitionManager.java
+++ b/src/java/com/twitter/search/earlybird/archive/ArchiveSearchPartitionManager.java
@@ -326,19 +326,19 @@ public class ArchiveSearchPartitionManager extends PartitionManager {
 
     final ArchiveSegment oldSegment = (ArchiveSegment) segmentInfo.getSegment();
 
-    return indexSegment(newSegmentInfoForIndexing, oldSegmentInfo, input -> {
+    return indexSegment(newSegmentInfoForIndexing, oldSegmentInfo, input ->
       // we're updating the segment - only index days after the old end date, but only if
       // we're in the on-disk archive, and we're sure that the previous days have already
       // been indexed.
-      return !earlybirdIndexConfig.isIndexStoredOnDisk()
+       !earlybirdIndexConfig.isIndexStoredOnDisk()
           // First time around, and the segment has not been indexed and optimized yet,
           // we will want to add all the days
           || !oldSegmentInfo.isOptimized()
           || oldSegmentInfo.getIndexSegment().getIndexStats().getStatusCount() == 0
           || !oldSegment.getDataEndDate().before(timeSlice.getEndDate())
           // Index any new days
-          || input.after(oldSegment.getDataEndDate());
-    });
+          || input.after(oldSegment.getDataEndDate())
+    );
   }
 
   private boolean existingSegmentNeedsUpdating(ArchiveTimeSlice timeSlice,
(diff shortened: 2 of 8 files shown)


Codfish246:
efficient

src/java/com/twitter/search/core/earlybird/facets/AbstractFacetCountingArray.java
```diff
@@ -93,8 +93,7 @@ public void collect(int docID) {
 
   protected static final int decodeTermID(int facetID) {
```
bageshwar, line 94: Dont understand the reason for this method to be a `protected static final`

- if its static, what's the point of protected
- if its protected what's the point of final

src/java/com/twitter/search/earlybird/archive/ArchiveSearchPartitionManager.java
```diff
@@ -326,19 +326,19 @@ private boolean indexNewDayAndAppendExistingSegment(final ArchiveTimeSlice timeS
...
           // First time around, and the segment has not been indexed and optimized yet,
           // we will want to add all the days
           || !oldSegmentInfo.isOptimized()
           || oldSegmentInfo.getIndexSegment().getIndexStats().getStatusCount() == 0
           || !oldSegment.getDataEndDate().before(timeSlice.getEndDate())
           // Index any new days
-          || input.after(oldSegment.getDataEndDate());
-    });
+          || input.after(oldSegment.getDataEndDate())
+    );
```
pedroluiznogueira, line 341: IMHO, I dunno about that, maybe this lambda looks better with the `{}` since it has a bunch of comments inside of it.

src/java/com/twitter/search/core/earlybird/facets/AbstractFacetCountingArray.java
pedroluiznogueira: on the second point, protected only wouldn't ensure it can't be overriden?

KingParity:
💀 the engineers here be slackin

src/java/com/twitter/search/earlybird/archive/ArchiveHDFSUtils.java
```diff
@@ -24,7 +24,7 @@ public final class ArchiveHDFSUtils {
   private static final Logger LOG = LoggerFactory.getLogger(ArchiveHDFSUtils.class);
 
   private static final Pattern SEGMENT_NAME_PATTERN =
-      Pattern.compile("_start_([0-9]+)_p_([0-9]+)_of_([0-9]+)_([0-9]{14}+)_");
+      Pattern.compile("_start_(\\d+)_p_(\\d+)_of_(\\d+)_(\\d{14}+)_");
```
iashchak, line 27: Actually it is not the same.
१२३४५६७८९० as example passes \d but not 0-9 pattern

laoal11: you are right, I changed it back

src/java/com/twitter/search/earlybird/search/AbstractResultsCollector.java
```diff
@@ -193,11 +193,10 @@ private static FacetLabelCollector buildFacetCollector(
       }
     }
 
-    if (requiredFields.size() > 0) {
-      return new FacetLabelCollector(requiredFields);
-    } else {
+    if (requiredFields.size().isEmpty()) {
```
tal-yac, line 196: Supposed to be `requiredFields.isEmpty()` maybe?

Also I feel like a ternary operator would be a better fit, something like:
`return requiredFields.isEmpty() ? null :  new FacetLabelCollector(requiredFields);`

laoal11: good catch and I agree, changed it

src/java/com/twitter/search/earlybird/archive/ArchiveSearchPartitionManager.java
```diff
@@ -326,19 +326,19 @@ private boolean indexNewDayAndAppendExistingSegment(final ArchiveTimeSlice timeS
 
     final ArchiveSegment oldSegment = (ArchiveSegment) segmentInfo.getSegment();
 
-    return indexSegment(newSegmentInfoForIndexing, oldSegmentInfo, input -> {
+    return indexSegment(newSegmentInfoForIndexing, oldSegmentInfo, input ->
       // we're updating the segment - only index days after the old end date, but only if
       // we're in the on-disk archive, and we're sure that the previous days have already
       // been indexed.
-      return !earlybirdIndexConfig.isIndexStoredOnDisk()
+       !earlybirdIndexConfig.isIndexStoredOnDisk()
```
Taz03, line 333:
```suggestion
      !earlybirdIndexConfig.isIndexStoredOnDisk()
```
fix indentation

src/java/com/twitter/search/earlybird/archive/ArchiveSearchPartitionManager.java
Taz03: i mean it still looks the same so...

PR #572, opened 2023-04-01
Fix 10k favs check

marinight:
It originally was check if it was greater than or equal to 1000. This commit would fix the issue by replacing 1000 with the correct number, 10000.

### DIFF
--- a/home-mixer/server/src/main/scala/com/twitter/home_mixer/functional_component/decorator/HomeTweetTypePredicates.scala
+++ b/home-mixer/server/src/main/scala/com/twitter/home_mixer/functional_component/decorator/HomeTweetTypePredicates.scala
@@ -160,7 +160,7 @@ object HomeTweetTypePredicates {
     ("has_gte_1k_favs", _.getOrElse(EarlybirdFeature, None).exists(_.favCountV2.exists(_ >= 1000))),
     (
       "has_gte_10k_favs",
-      _.getOrElse(EarlybirdFeature, None).exists(_.favCountV2.exists(_ >= 1000))),
+      _.getOrElse(EarlybirdFeature, None).exists(_.favCountV2.exists(_ >= 10000))),
     (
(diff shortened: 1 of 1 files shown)


Codfish246 (commented):
lgtm

looskie (commented):
LGTM

hexiro:
LGTM!

hexiro (commented):
LGTM

victornery:
LGTM

home-mixer/server/src/main/scala/com/twitter/home_mixer/functional_component/decorator/HomeTweetTypePredicates.scala
```diff
@@ -160,7 +160,7 @@ object HomeTweetTypePredicates {
     ("has_gte_1k_favs", _.getOrElse(EarlybirdFeature, None).exists(_.favCountV2.exists(_ >= 1000))),
     (
       "has_gte_10k_favs",
-      _.getOrElse(EarlybirdFeature, None).exists(_.favCountV2.exists(_ >= 1000))),
+      _.getOrElse(EarlybirdFeature, None).exists(_.favCountV2.exists(_ >= 10000))),
```
nikolaydubina, line 163: omg twitter code is 🤡-town

AnEmortalKid: Use 10_000 to avoid clowning ourselves in the future

AnEmortalKid: Also add a test that would expose this bug.

AlexFlipnote:
I was actually expecting a joke PR, but seems to actually fix something 🙃

PR #638, opened 2023-04-01
fix(libtwml): use fstream in BlockFormatWriter

ElijahPepe:
Tested on Ubuntu 22.04.

### DIFF
--- a/twml/libtwml/src/lib/BlockFormatWriter.cpp
+++ b/twml/libtwml/src/lib/BlockFormatWriter.cpp
@@ -11,7 +11,7 @@
 #endif
 
 #define MARKER_SIZE (16)
-static uint8_t _marker[MARKER_SIZE] = {
+static std::uint8_t _marker[MARKER_SIZE] = {
         0x29, 0xd8, 0xd5, 0x06, 0x58, 0xcd, 0x4c, 0x29,
         0xb2, 0xbc, 0x57, 0x99, 0x21, 0x71, 0xbd, 0xff
 };
@@ -21,10 +21,13 @@ namespace twml {
       file_name_(file_name), record_index_(0), records_per_block_(record_per_block) {
       snprintf(temp_file_name_, PATH_MAX, "%s.block", file_name);
       std::fstream outputfile_(file_name_, std::ios::app);
+      if (!outputfile_.is_open()) {
+        throw std::runtime_error("Failed to open the output file");
+      }
     }
 
-    int BlockFormatWriter::pack_tag_and_wiretype(std::fstream& fs, uint32_t tag, uint32_t wiretype) {
-      uint8_t x = ((tag & 0x0f) << 3) | (wiretype & 0x7);
+    int BlockFormatWriter::pack_tag_and_wiretype(std::fstream& fs, std::uint32_t tag, std::uint32_t wiretype) {
+      std::uint8_t x = ((tag & 0x0f) << 3) | (wiretype & 0x7);
       fs.write(reinterpret_cast<char*>(&x), sizeof(x));
       if (fs.fail()) {
         return -1;
@@ -34,10 +37,10 @@ namespace twml {
 
     int BlockFormatWriter::pack_varint_i32(std::fstream& fs, int value) {
       for (int i = 0; i < 10; i++) {
-        uint8_t x = value & 0x7F;
+        std::uint8_t x = value & 0x7F;
         value = value >> 7;
         if (value != 0) x |= 0x80;
-        fs.write(reinterpret_cast<char*>(&x), sizeof(x));
+        fs.write(reinterpret_cast<const char*>(&x), sizeof(x));
         if (fs.fail()) {
           return -1;
         }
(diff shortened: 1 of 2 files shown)


twml/libtwml/src/lib/BlockFormatWriter.cpp
```diff
@@ -74,9 +72,9 @@ namespace twml {
       if (record) {
         record_index_++;
         // The buffer holds max records_per_block_ of records (block).
-        FILE *buffer = fopen(temp_file_name_, "a");
+        std::fstream buffer(temp_file_name_, std::ios::app);
         if (!buffer) return -1;
```
bwedding, line 76: std::fstream constructor does not return null on fail. You need to check is_open()

`if (!buffer.is_open()) return -1;`

twml/libtwml/src/lib/BlockFormatWriter.cpp
```diff
@@ -20,51 +20,49 @@ namespace twml {
...
+        return -1;
+      }
       return 0;
     }
 
-    int BlockFormatWriter::write_int(FILE *buffer, int value) {
+    int BlockFormatWriter::write_int(std::fstream& fs, int value) {
       uint8_t buff[4];
       buff[0] = value & 0xff;
       buff[1] = (value >> 8) & 0xff;
```
bwedding, line 61: Why use C-style casts? It's CPP and you've already used a reinterpret cast. These should be:
`buff[1] = static_cast<std::uint8_t>((value >> 8) & 0xff);`

twml/libtwml/src/lib/BlockFormatWriter.cpp
```diff
@@ -20,51 +20,49 @@ namespace twml {
...
+      fs.write(in, in_len);
+      if (fs.fail()) {
+        return -1;
+      }
       return 0;
     }
 
-    int BlockFormatWriter::write_int(FILE *buffer, int value) {
+    int BlockFormatWriter::write_int(std::fstream& fs, int value) {
       uint8_t buff[4];
```
bwedding, line 59: std::uint8_t is more portable

twml/libtwml/src/lib/BlockFormatWriter.cpp
```diff
@@ -101,24 +98,24 @@ namespace twml {
...
         if (write_int(outputfile_, block_size)) return 1;
         uint8_t buff[4096];
         while (1) {
-          size_t n = fread(buff, 1, sizeof(buff), buffer);
+          buffer.read(reinterpret_cast<char*>(buff), sizeof(buff));
+          std::streamsize n = buffer.gcount();
           if (n) {
-            size_t x = fwrite(buff, 1, n, outputfile_);
-            if (x != n) return 1;
+            outputfile_.write(reinterpret_cast<char*>(buff), n);
```
bwedding, line 114: Should be const:

`outputfile_.write(reinterpret_cast<const char*>(buff)`

twml/libtwml/src/lib/BlockFormatWriter.cpp
```diff
@@ -20,51 +20,49 @@ namespace twml {
     BlockFormatWriter::BlockFormatWriter(const char *file_name, int record_per_block) :
       file_name_(file_name), record_index_(0), records_per_block_(record_per_block) {
       snprintf(temp_file_name_, PATH_MAX, "%s.block", file_name);
-      outputfile_ = fopen(file_name_, "a");
+      std::fstream outputfile_(file_name_, std::ios::app);
```
bwedding, line 23: This needs to throw if it can't open the file (RAII). Add something like this to the constructor:

```
        if (!outputfile_.is_open()) {
            throw std::runtime_error("Failed to open the output file");
        }
```

twml/libtwml/src/lib/BlockFormatWriter.cpp
```diff
@@ -20,51 +20,49 @@ namespace twml {
...
 
-    int BlockFormatWriter::pack_varint_i32(FILE *buffer, int value) {
+    int BlockFormatWriter::pack_varint_i32(std::fstream& fs, int value) {
       for (int i = 0; i < 10; i++) {
         uint8_t x = value & 0x7F;
         value = value >> 7;
         if (value != 0) x |= 0x80;
-        size_t n = fwrite(&x, 1, 1, buffer);
-        if (n != 1) {
+        fs.write(reinterpret_cast<char*>(&x), sizeof(x));
```
bwedding, line 40: cast to const char*

twml/libtwml/src/lib/BlockFormatWriter.cpp
ElijahPepe: Thank you, I overlooked that converting it.

twml/libtwml/src/lib/BlockFormatWriter.cpp
ElijahPepe: Overlooked that, thank you!

twml/libtwml/src/lib/BlockFormatWriter.cpp
ElijahPepe: Overlooked that, thank you!

PR #647, opened 2023-04-01
feat: reduce duplication on BasicIndexingConverter and EncodedFeatureBuilder to improve readability and maintainability

pedroluiznogueira:
- as this method already has a java doc, why not have the entire logic explained on it, instead of inside the method so it doesn't make it hard to read the actual code

- since the lambda parameters of the `Predicate` represent an id, giving them better names than 'x', and extracting the `Predicate` as a local variable to avoid duplication

- use `Optional.ifPresent` instead of an `if Optional.ifPresent`, not only to make a better use of Optionals but also to reduce duplication of having to `get` the optional value


Ghurtchu:
LGTM 🚀

pedroluiznogueira:
can we get some reviews in here? @twitter-service

PR #652, opened 2023-04-01
perf: Optimize code for initializing Kafka consumers and processors

iamrajiv:
## Description

Refactor the code in `initRecosHoseKafka` to simplify and optimize it, while maintaining its functionality. Changes include:

Here's a breakdown of the changes made:

- Defining **`processors`** and **`consumers`** variables inside the **`try`** block:
  By defining these variables inside the **`try`** block, we avoid polluting the class scope with variables that are only used inside this function. This also makes it easier to reason about the scope and lifetime of these variables.

- Using **`map`** instead of **`foreach`** to create **`consumers`** and **`processors`**:
  Using **`map`** to create the **`consumers`** and **`processors`** makes the code more concise and easier to read. It also allows us to store the results of these operations in variables, which can be useful if we need to reference them later in the code.

- Method chaining with **`foreach`**:
  Instead of using a separate statement to call **`start()`** on each **`processor`**, we can use method chaining with **`foreach`** to make the code more concise.

- Removing unnecessary code:
  The statements **`processors = Seq.empty`** and **`consumers = Seq.empty`** inside the **`catch`** block are unnecessary since the variables will be redefined the next time the function is called. Removing them simplifies the code and makes it easier to read.

### DIFF
--- a/src/scala/com/twitter/recos/hose/common/UnifiedGraphWriter.scala
+++ b/src/scala/com/twitter/recos/hose/common/UnifiedGraphWriter.scala
@@ -85,50 +85,40 @@ trait UnifiedGraphWriter[
     }
   }
 
-  private def initRecosHoseKafka(
-    queue: java.util.Queue[Array[RecosHoseMessage]],
-    queuelimit: Semaphore,
-  ): Unit = {
-    try {
-      consumers = (0 until consumerNum).map { index =>
-        new ThreadSafeKafkaConsumerClient(
-          kafkaConsumerBuilder.clientId(s"clientId-$index").enableAutoCommit(false).config)
-      }
-      processors = consumers.zipWithIndex.map {
-        case (consumer, index) =>
-          val bufferedWriter = BufferedEdgeCollector(bufferSize, queue, queuelimit, statsReceiver)
-          val processor = RecosEdgeProcessor(bufferedWriter)(statsReceiver)
-
-          AtLeastOnceProcessor[String, RecosHoseMessage](
-            s"recos-injector-kafka-$index",
-            hosename,
-            consumer,
-            processor.process,
-            maxPendingRequests = MaxPendingRequests * bufferSize,
-            workerThreads = ProcessorThreads,
-            commitIntervalMs = CommitIntervalMs,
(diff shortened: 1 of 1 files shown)


src/scala/com/twitter/recos/hose/common/UnifiedGraphWriter.scala
```diff
@@ -85,50 +85,40 @@ trait UnifiedGraphWriter[
...
-            hosename,
-            consumer,
-            processor.process,
-            maxPendingRequests = MaxPendingRequests * bufferSize,
-            workerThreads = ProcessorThreads,
-            commitIntervalMs = CommitIntervalMs,
-            statsReceiver = statsReceiver
-          )
-      }
+private def initRecosHoseKafka(queue: java.util.Queue[Array[RecosHoseMessage]], queuelimit: Semaphore): Unit = {
```
apal0934, line 88: The existing style should be preserved

src/scala/com/twitter/recos/hose/common/UnifiedGraphWriter.scala
```diff
@@ -85,50 +85,40 @@ trait UnifiedGraphWriter[
...
-        e.printStackTrace()
-        log.error(e, e.toString)
-        processors.foreach { processor =>
-          processor.close()
-        }
-        processors = Seq.empty
-        consumers.foreach { consumer =>
-          consumer.close()
-        }
-        consumers = Seq.empty
```
apal0934, line 129: processors and consumers aren't local variables, what happens if another function tries to use them after an exception in the future? They should be cleared for safety.

iamrajiv: @apal0934 I am using the `processors.foreach(_.close())` and `consumers.foreach(_.close())` methods to ensure that any resources held by the processors and consumers collections are properly released and can be reused by the system.

apal0934: But performance wise it's way faster to set them to nothing than to let another function think there's elements in the sequence and try to loop

PR #660, opened 2023-04-01
Limit penalization on blocks / mutes for a cooldown of 180 days. Fix #658

### LINKED ISSUE #658: Excessively penalizing accounts for being blocked or reported
Either you've experienced it or you're very lucky. Sometimes, even during civil conversation, someone will be in a bad mood and suddenly block you - even someone who you've regularly interacted with!

Please reduce the penalty for this. Also, shadowbanning and/or reducing account "reach" silently, without transparency, is wrong (but much appreciation for this major step towards transparency). Please show us exactly what we've done to be deboosted. That way, we can correct the behavior and remain allowed into the public square.

Please indicate which of our posts have been reported so that we can correct the behavior.

It's truly frustrating (and Orwellian) when a previously well-functioning account is suddenly extremely throttled. Please increase the speed at which an account can recover after being sent to the e-gulag.

<3 Musk and the new twitter. You're the good guys, and we all know it! You're saving democracy and so much more. Free speech is the bedrock of everything. "Freedom of speech, not freedom of reach" is the antithesis of free speech, however.

syrusakbary:
Currently, is reported that blocks and mutes are excessively penalized (see #658).

I myself experimented the side effects from cancellation efforts from a woke majority and saw how engagement decreased significantly after that happened.
By setting a window limit on blocks and mutes we still take those blocks in consideration, but only during a cooldown period (180 days), so the account is not permanently cancelled.

This PR fixes it by setting a cooldown of 180 days for blocks and mutes

### DIFF
--- a/src/scala/com/twitter/interaction_graph/scio/agg_negative/InteractionGraphNegativeJob.scala
+++ b/src/scala/com/twitter/interaction_graph/scio/agg_negative/InteractionGraphNegativeJob.scala
@@ -49,17 +49,30 @@ object InteractionGraphNegativeJob extends ScioBeamJob[InteractionGraphNegativeO
     val endTs = opts.interval.getEndMillis
 
     // read input datasets
+
+    // We only count blocks in the past 180 days to make sure we don't fall
+    // into any long-term cancellation effects in a tweeter.
+    // As an example, some accounts may have higher count of blocks because
+    // what they said is controversial in one period of time but it might be
+    // not in other (for example, @RWMaloneMD or others)
     val blocks: SCollection[InteractionGraphRawInput] =
       GraphUtil.getFlockFeatures(
         readSnapshot(FlockBlocksEdgesScalaDataset, sc),
         FeatureName.NumBlocks,
         endTs)
+      .filter(_.age < 180)
 
+    // We only count mutes in the past 180 days to make sure we don't fall
+    // into any long-term cancellation effects in a tweeter.
+    // As an example, some accounts may have higher count of blocks because
+    // what they said is controversial in one period of time but it might be
+    // not in other (for example, @RWMaloneMD or others)
     val mutes: SCollection[InteractionGraphRawInput] =
       GraphUtil.getFlockFeatures(
         readSnapshot(FlockMutesEdgesScalaDataset, sc),
         FeatureName.NumMutes,
         endTs)
+      .filter(_.age < 180)
 
     val abuseReports: SCollection[InteractionGraphRawInput] =
       GraphUtil.getFlockFeatures(


src/scala/com/twitter/interaction_graph/scio/agg_negative/InteractionGraphNegativeJob.scala
```diff
@@ -49,17 +49,30 @@ object InteractionGraphNegativeJob extends ScioBeamJob[InteractionGraphNegativeO
     val endTs = opts.interval.getEndMillis
 
     // read input datasets
+
+    // We only count blocks in the past 180 days to make sure we don't fall
+    // into any long-term cancellation effects in a tweeter.
+    // As an example, some accounts may have higher count of blocks because
+    // what they said is controversial in one period of time but it might be
+    // not in other (for example, @RWMaloneMD or others)
```
nikolaydubina, line 57: good point

kadeksuryam:
LGTM!

setlightlyupon:
Great! I wonder, would 180 days still be an excessively long time? I suspect twitter will likely decide on their own duration, in any case.

khatharr:
> Great! I wonder, would 180 days still be an excessively long time? I suspect twitter will likely decide on their own duration, in any case.

Right.
https://en.wikipedia.org/wiki/Magic_number_(programming)

A simple cutoff may not be the best method either. Having all stats decay by some function over time may be a better idea, though it might be expensive in terms of compute time, since the docs say that the scio graph is generated several times a day. 

When I follow the trail of function calls here, I see that this is being used to generate a graph of various statistics between users that gets fed into the AI in order to try and predict who will interact with who. I'm not seeing a place anywhere that states explicitly that blocks and mutes are "penalized", only that they're offered to the robot overlord as a metric for consideration. It may well be that some users are more likely to pursue interactions with users who have a lot of blocks, while others don't, and the bot is trying to sort out who's who. It's probably worth trying to untangle the knot a bit more.

pol0nium:
Maybe an approach similar to the [HN](news.ycombinator.com) algorithm[(1)](https://medium.com/hacking-and-gonzo/how-hacker-news-ranking-algorithm-works-1d9b0cf2c08d) (I think Reddit partially uses the same concept) could be smoother.

Some weights like "blocks" and "mutes" could be decreased by a factor of time and gravity.

Basically, the older it is, the less it counts.

izzy:
> I myself experimented the side effects from cancellation efforts from a woke majority and saw how engagement decreased significantly after that happened.

So you said or did something and now you don't want to live with the consequences of people not wanting to talk to you.

I have a better fix for this: Grow up instead of trying to commit changes so you don't have to learn.

syrusakbary:
> I have a better fix for this: Grow up instead of trying to commit changes so you don't have to learn

@izzy I indeed grew from it… the algorithm didn’t 😉

khatharr:
> Some weights like "blocks" and "mutes" could be decreased by a factor of time and gravity.

Diminishing only the subset of metrics that humans feel negatively about will only cause inflation of the other metrics. If the NN is any good then it will just inflate the reduced metrics to compensate, meaning that recent blocks/mutes would likely have oversized effects compared to the present.

Nightcaat:
LGTM

Nightcaat (commented):
LGTM

wolfvandenbussche:
I actively fight against the full-cancellation of tweeters. LGTM

jamesdigid:
LGTM, simple and to the point and if anything a good start.

redknightlois:
It may be a good stop-gap method, but this does not solve the problem. On this comment I layout how you would attack this time-based decay. Just unblock and reblock the account would essentially reset the counter. https://github.com/twitter/the-algorithm/issues/1386#issuecomment-1499395522

jamesdigid:
Twitter doesn't have a cool off period for block/re-block? If not that's a separate issue entirely and doesn't negate this one.

amark:
Using blocks should punish the blocker, not audiences. There's no better way to suppress whistleblowers than giving rings the ability to silence dissent. Twitter had a proactive goal of facilitating "healthy public discourse", blocks are the opposite of that (tho fine for personal reasons).

My small/moderate account (5.5K followers) went from ~2K visibility/tweet down to 200 after getting blocked by Yud (AI doom guy), I don't block anyone. It is really discouraging and makes using Twitter near pointless if *my followers* aren't actually seeing my content as they've explicitly opted into - they're not really mine.

Next up: I start an email newsletter? :/ Let's get these changes merged. Tho 180 days is also insane, blocks should have 0 effect on others, only what you see.

syrusakbary:
Hoping to see this merged at some time!

Naim2000:
> LGTM

@Nightcaat What are you doing here

PR #664, opened 2023-04-01
perf: Improve performance of DocValuesHelper class

iamrajiv:
## Description

The **`DocValuesHelper`** class in the **`com.twitter.search.earlybird.index`** package has been optimized for performance and readability. Specifically, the following changes have been made:

- The **`LeafReader`** object is now handled using a try-with-resources block. This ensures that the object is closed properly after use, preventing resource leaks and simplifying the code.
- The **`TermsEnum`** object is now cached using a **`ConcurrentHashMap`**. This avoids creating a new object every time **`getFirstDocIdWithValue()`** and **`getLargestDocIdWithCeilOfValue()`** are called. Caching the **`TermsEnum`** object reduces the overhead of object creation and can improve the performance of the code.
- **`DocIdSetIterator.NO_MORE_DOCS`** is now used instead of **`1`** to indicate the end of a search. Using **`DocIdSetIterator.NO_MORE_DOCS`** can make the code more readable and self-explanatory.

These changes can improve the performance of the code and make it easier to read and maintain.

### DIFF
--- a/src/java/com/twitter/search/earlybird/index/DocValuesHelper.java
+++ b/src/java/com/twitter/search/earlybird/index/DocValuesHelper.java
@@ -1,6 +1,8 @@
 package com.twitter.search.earlybird.index;
 
 import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.lucene.index.LeafReader;
 import org.apache.lucene.index.Terms;
@@ -9,6 +11,8 @@ import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.util.BytesRef;
 
 public final class DocValuesHelper {
+  private static final Map<String, TermsEnum> termsEnumCache = new ConcurrentHashMap<>();
+
   private DocValuesHelper() {
   }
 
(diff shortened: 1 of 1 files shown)


src/java/com/twitter/search/earlybird/index/DocValuesHelper.java
```diff
@@ -22,13 +26,10 @@ private DocValuesHelper() {
    */
   public static int getFirstDocIdWithValue(
       LeafReader reader, String indexField, BytesRef value) throws IOException {
-    TermsEnum termsEnum = getTermsEnum(reader, indexField);
-    if (termsEnum == null || !termsEnum.seekExact(value)) {
-      return DocIdSetIterator.NO_MORE_DOCS;
+    try (DocIdSetIterator docsIterator = getTermsEnum(reader, indexField).postings(null)) {
+      int docId = docsIterator.nextDoc();
+      return docId == DocIdSetIterator.NO_MORE_DOCS ? docId : docId;
```
kevinam99, line 31: huh?

iamrajiv: @kevinam99 The change made to the **`getFirstDocIdWithValue()`** method is to use a try-with-resources block to handle the **`DocIdSetIterator`** object returned by **`termsEnum.postings(null)`** method call. This ensures that the **`DocIdSetIterator`** object is closed properly after use, avoiding potential resource leaks and simplifying the code.

xxDark: You are potentially exposing the same `TermsEnum` to multiple threads that have a different `LeafReader` instance but use the same `indexField`. This is not an expected behavior and will result in errors.

PR #670, opened 2023-04-01
Rename tweepcred README to README.md

apal0934:
This allows GitHub and other tools to correctly display the Markdown.


kapilepatel:
You will need to also update any file where this is linked like https://github.com/twitter/the-algorithm/blob/main/README.md

PR #687, opened 2023-04-01
feat: do not generate a stream that will be processed right after

pedroluiznogueira:


AnEmortalKid (commented):
Neat, uses the method from the Interable directly https://docs.oracle.com/javase/8/docs/api/java/lang/Iterable.html#forEach-java.util.function.Consumer-

kapilepatel (commented):
looks good

PR #706, opened 2023-04-01
Important Twitter Fix: Unionization

mitDebo:
Every worker deserves a union. That includes the workers at twitter. If you would like information on how to start organizing your workplace, or if you need assistance in your organizing, contact the IWW: https://www.iww.org/organize/


rd-99:
LGTM

PR #710, opened 2023-04-01
Format

zihaoAK47:
Bad annotation. If this class is exposed, please provide a good annotation format

### DIFF
--- a/src/java/com/twitter/search/core/earlybird/index/util/SearchSortUtils.java
+++ b/src/java/com/twitter/search/core/earlybird/index/util/SearchSortUtils.java
@@ -21,7 +21,7 @@ public abstract class SearchSortUtils {
     int high = end;
     Preconditions.checkState(comparator.compare(low, key) <= comparator.compare(high, key));
     while (low <= high) {
-      int mid = (low + high) >>> 1;
+      int mid = (low & high) + ((low ^ high) >>> 1);
       int result = comparator.compare(mid, key);
       if (result < 0) {
         low = mid + 1;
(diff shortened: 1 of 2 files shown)


src/java/com/twitter/search/core/earlybird/index/util/SearchSortUtils.java
```diff
@@ -21,7 +21,7 @@ public static <T> int binarySearch(Comparator<T> comparator, final int begin, fi
     int high = end;
     Preconditions.checkState(comparator.compare(low, key) <= comparator.compare(high, key));
     while (low <= high) {
-      int mid = (low + high) >>> 1;
+      int mid = (low & high) + ((low ^ high) >>> 1);
```
kapilepatel, line 24: Can you explain this?

zihaoAK47: int mid = (low + high) >>> 1;
Of course, the above operations may cause overflow. It's just possible.

int mid = (low & high) + ((low ^ high) >> 1);
This operation can ensure no overflow and correctly calculate the middle position of low and high.

zihaoAK47: int mid = (low + high) >>> 1;
Of course, data overflow may occur in the above situation.
You may have verified that the result is correct, but there was a data overflow during the calculation process.

int mid = (low & high) + ((low ^ high) >>> 1);
This method is bit wise and does not cause data overflow Whether it is the result or the calculation process.

PR #716, opened 2023-04-01
Add CodeQL and Codacy security scanning

wreiske:
This PR adds support for CodeQL and Codacy security scanning through GitHub Actions.

Screenshot 2023-03-31 at 11 05 54 PM

### DIFF
--- a/.github/workflows/codacy.yml
+++ b/.github/workflows/codacy.yml
@@ -1,0 +1,61 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+# This workflow checks out code, performs a Codacy security scan
+# and integrates the results with the
+# GitHub Advanced Security code scanning feature.  For more information on
+# the Codacy security scan action usage and parameters, see
+# https://github.com/codacy/codacy-analysis-cli-action.
+# For more information on Codacy Analysis CLI in general, see
+# https://github.com/codacy/codacy-analysis-cli.
+
+name: Codacy Security Scan
+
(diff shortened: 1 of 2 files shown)


HarryZ10:
Have you verified that these can be resolved? They can surely bring up 21,000 alerts, but I'm curious which are the urgent ones to incorporate a Github Action. Like in the screenshot, maybe there is a good reason for an "overly permissive regex range"

wreiske:
> Have you verified that these can be resolved? They can surely bring up 21,000 alerts, but I'm curious which are the urgent ones to incorporate a Github Action. Like in the screenshot, maybe there is a good reason for an "overly permissive regex range"

At first glance, it looks like many of the Codacy alerts are lint related - for example, tabs out of place, or 4 spaces instead of 2 when all other tabs in the file were 2. Some of these could be fixed in batch by just formatting all the code to whatever their standard is or removing those types of alerts altogether from the scans.

The CodeQL alerts only found 9 things, all related to the EMOJI_PATTERN here: https://github.com/wreiske/the-algorithm/blob/78fcd2c2858fd5864a4dc84c0187bd1d765e964f/trust_and_safety_models/nsfw/nsfw_text.py#L42-L54

Screenshot 2023-03-31 at 11 26 18 PM

I didn't configure CodeQL to scan any of the Java or C++ because it requires building steps for each. It looks like they have removed some of the internal build steps, like https://github.com/wreiske/the-algorithm/blob/78fcd2c2858fd5864a4dc84c0187bd1d765e964f/ci/ci.sh that is completely empty. 

tl;dr - probably can remove a lot of the "formatting" lint checks and reduce the alerts significantly.

PR #790, opened 2023-04-01
Reduce visibility in default recent tweet seconds constant

DanielGilB:


src/java/com/twitter/search/earlybird/RecentTweetRestriction.java
```diff
@@ -11,7 +11,7 @@ public final class RecentTweetRestriction {
   private static final String QUERY_CACHE_UNTIL_TIME = "query_cache_until_time";
 
   @VisibleForTesting
-  public static final int DEFAULT_RECENT_TWEET_SECONDS = 15;
+  private static final int DEFAULT_RECENT_TWEET_SECONDS = 15;
```
kapilepatel, line 14: It might break in testing if it is made private, how did you confirm that it will not break?

DanielGilB: It has VisibleForTesting annotation

PR #826, opened 2023-04-01
Repo Getting started instructions

srohatgi01:
- Added some base line instructions about repository which can be extended further as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,23 @@
 # Twitter Recommendation Algorithm
 
 The Twitter Recommendation Algorithm is a set of services and jobs that are responsible for constructing and serving the
-Home Timeline. For an introduction to how the algorithm works, please refer to our [engineering blog](https://blog.twitter.com/engineering/en_us/topics/open-source/2023/twitter-recommendation-algorithm). The
-diagram below illustrates how major services and jobs interconnect.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,8 +1,23 @@
...
-Home Timeline. For an introduction to how the algorithm works, please refer to our [engineering blog](https://blog.twitter.com/engineering/en_us/topics/open-source/2023/twitter-recommendation-algorithm). The
-diagram below illustrates how major services and jobs interconnect.
+Home Timeline. For an introduction to how the algorithm works, please refer to our [engineering blog](https://blog.twitter.com/engineering/en_us/topics/open-source/2023/twitter-recommendation-algorithm). 
+
+Reading the Documentation
+-------------------------
+
+Initial documentation to get you started:
+
+- `COPYING`: The Twitter Recommendation Algorithm license agreement (AGPL-3.0 license).
```
ElijahPepe, line 11: The license is not important to the average contributor.

srohatgi01: True but it's good to describe what the document is in README.md as that's the first thing any contributor is going to read. Additionally my intention was that we can have contributing guidelines or something similar listed here so that it's easier for any new contributor to navigate.

PR #834, opened 2023-04-01
🔥🔥 FIXED TODO Replace score field with FeatureMap in TweetAnnotationQueryCandidate 🔥🔥

altruong:
TODO Remove score from the candidate and use a Feature instead

- Remove the score field from the TweetAnnotationQueryCandidate class
- Add a features field of type FeatureMap to the TweetAnnotationQueryCandidate class
- Update the apply method in the companion object of TweetAnnotationQueryCandidate to accept a FeatureMap
- Update the equals and hashCode methods in the TweetAnnotationQueryCandidate class to exclude the score field and include the features field
- Update any relevant tests or usage of TweetAnnotationQueryCandidate to include the new FeatureMap instead of the score

### DIFF
--- a/product-mixer/component-library/src/main/scala/com/twitter/product_mixer/component_library/model/candidate/suggestion/QuerySuggestionCandidate.scala
+++ b/product-mixer/component-library/src/main/scala/com/twitter/product_mixer/component_library/model/candidate/suggestion/QuerySuggestionCandidate.scala
@@ -231,11 +231,10 @@ object TypeaheadEventCandidate {
 /**
  * Canonical TweetAnnotationQueryCandidate model. Always prefer this version over all other variants.
  *
- * TODO Remove score from the candidate and use a Feature instead
  */
 final class TweetAnnotationQueryCandidate private (
   override val id: String,
-  val score: Option[Double])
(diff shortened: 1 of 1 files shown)


product-mixer/component-library/src/main/scala/com/twitter/product_mixer/component_library/model/candidate/suggestion/QuerySuggestionCandidate.scala
```diff
@@ -261,7 +260,8 @@ final class TweetAnnotationQueryCandidate private (
         (
           (this eq candidate)
             || ((hashCode == candidate.hashCode)
-              && (id == candidate.id && score == candidate.score))
+              && (id == candidate.id))
```
ImperiumTakp, line 263: this is a lot more clean then before

artonh: absolutely

artonh (commented):
Great work on this code! I really appreciate how you've used clear variable names. This is a well-written and efficient piece of code.

PR #839, opened 2023-04-01
The import re statement is duplicated

ImPavloh:
The second `import re` statement is redundant because it's already imported earlier


bigint (commented):
**LGTM** :shipit: 

Rad LGTM

PR #843, opened 2023-04-01
Refactor IdentifiableQueryScorer and offensive tweet classifications

aevitas:
The code defines a class called IdentifiableQueryScorer which extends the FilteredScorer class. It takes a weight, an inner scorer, a query ID, and an attribute collector as parameters. The constructor initializes these fields and calls the constructor of the superclass. 

The class overrides the iterator() method of the superclass to return a new DocIdSetIterator. This iterator wraps the iterator of the superclass, and on each call to nextDoc() or advance(), it collects the query ID and the scorer attribution for the current document using the attribute collector passed to the constructor. Finally, it returns the document ID. 

In summary, the IdentifiableQueryScorer class provides a way to identify which query a document matches using an attribute collector, and it does so by wrapping the iterator of the superclass and collecting the relevant information on each call to nextDoc() or advance().

I've refactored both the IdentifiableQueryScorer as well as the QueryRankVisitor to be in line with the following basic process:

1. An `IdentityHashMap<Query, Integer>` called `nodeToRankMap` is used to store the relationship between query objects and their node ranks.

2. The constructor `QueryRankVisitor()` initializes the superclass `DetectAnnotationVisitor` with the `Annotation.Type.NODE_RANK` type.

3. `visitBooleanQuery(BooleanQuery query)` method:
   - This method takes a `BooleanQuery` object as input and checks if it has an annotation of type `NODE_RANK`.
   - If it has the `NODE_RANK` annotation, it calls the `collectNodeRank()` method to store the node rank in the `nodeToRankMap`.
   - It then iterates through the children of the `BooleanQuery` object, and returns true if at least one child accepts the visitor, otherwise false.

4. `visitQuery(Query query)` method:
   - This method takes a `Query` object as input and checks if it has an annotation of type `NODE_RANK`.
   - If it has the `NODE_RANK` annotation, it calls the `collectNodeRank()` method to store the node rank in the `nodeToRankMap`, and returns true.
   - If the `Query` object does not have a `NODE_RANK` annotation, the method returns false.

5. `collectNodeRank(Annotation anno, Query query)` method:
   - This private method takes an `Annotation` object and a `Query` object as input.
   - It checks if the annotation type is `NODE_RANK`, extracts the rank value (as an integer) from the annotation, and stores it in the `nodeToRankMap` with the query object as the key.

6. `getNodeToRankMap()` method:
   - This public method returns the `nodeToRankMap`, allowing other classes to access the collected node ranks.

In summary, the `QueryRankVisitor` class is used to traverse query objects and collect their node ranks (if present) in an `IdentityHashMap`. This can be useful in scenarios where the rank information of query nodes needs to be accessed or processed later.

Overall this should improve the query and tweet suggestion behavior, especially in marginal cases such as the delimited historical tweets which have been pseudo-archived. 

Thanks for considering and of course I'm here to answer any and all questions.


aevitas:
Thanks.

PR #855, opened 2023-04-01
Update README.md

shreyanshhub:
Updated wikipedia links for cosine similarity,dot product and Linear algebra


zer0-x:
Shouldn't you use `en.wikipedia.org` rather then `en.m.wikipedia.org`?

The `en.m.wikipedia.org` domain is used for mobile phones (small screens), while the normal one is for desktops.

shreyanshhub:
> Shouldn't you use `en.wikipedia.org` rather then `en.m.wikipedia.org`?
> 
> The `en.m.wikipedia.org` domain is used for mobile phones (small screens), while the normal one is for desktops.

Updated to en.wikipedia.org

PR #922, opened 2023-04-01
Fix f-string syntax error

marianovik:
### Description

This merge request includes several changes, such as:

- Fixed a syntax error in an f-string by removing a redundant '}'. [Go to the line.](https://github.com/twitter/the-algorithm/pull/922/files#diff-c54399ecdca21316030441c443dd7a639a9d8815469b2d040de6a31f470df197R431)
- Replaced print() statements with logging.info() to improve logging functionality.
- Changed the size parameter in `plt.title()` to `fontsize`, as `size` is not a valid parameter for `plt.title()`.

### References:

[PEP 282 -- A Logging System]( https://www.python.org/dev/peps/pep-0282/)
[Matplotlib Documentation
](https://matplotlib.org/stable/api/pyplot_api.html#matplotlib.pyplot.title)

### DIFF
--- a/trust_and_safety_models/nsfw/nsfw_media.py
+++ b/trust_and_safety_models/nsfw/nsfw_media.py
@@ -1,5 +1,6 @@
+import logging
+
 import kerastuner as kt
-import math
 import numpy as np
 import pandas as pd
 import random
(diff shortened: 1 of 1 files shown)


aaronmondal:
I'm in favor of improving the code quality of the repo, but I think reformatting and linting issues need to be handled with separate commits. Otherwise this gets hard to review and it becomes unclear which files are fixed and which aren't. We also need a single source of truth regarding style issues.

My personal favorite for Python is the [wemake-python-styleguide](https://github.com/wemake-services/wemake-python-styleguide) because that **actually** improves code quality noticeably beyond just formatting issues. We could enforce it in the pre-commit hooks in https://github.com/twitter/the-algorithm/pull/441.

aaronmondal:
Ah well it looks like Twitter already has a styleguide which we'll likely need to follow: https://github.com/twitter-archive/commons/blob/master/src/python/twitter/common/styleguide.md

marianovik:
@aaronmond Thank you for sharing your insights. I have updated the MR without making any changes to the formatting, as I agree with you that we should have a single source of truth.

I believe that it would be best to address all concerns separately in issues.

PR #923, opened 2023-04-01
Fix and optimize mixDiscreteIdAndValue() and mixStringIdAndValue() functions

wallbloggerbeing:
This pull request fixes and optimizes the mixDiscreteIdAndValue() and mixStringIdAndValue() functions in the twml namespace.

Specifically, the changes are:

-Added #include directives for standard library headers that are used.
-Removed the #ifdef __cplusplus wrapper, since it's unnecessary.
-Added const qualifier to str argument in mixStringIdAndValue() function to indicate that the function does not modify the contents of str.
-Replaced the for loop with a while loop in mixStringIdAndValue(), which avoids the overhead of computing i on each iteration and allows us to use a pointer instead of an index.
-Added braces around the body of both functions, which makes the code easier to read and less error-prone.
-Added a comment indicating the end of the twml namespace.

These changes should make the code more efficient and easier to read. Please review and let me know if you have any feedback. Thank you!

### DIFF
--- a/twml/libtwml/include/twml/utilities.h
+++ b/twml/libtwml/include/twml/utilities.h
@@ -1,18 +1,21 @@
 #pragma once
-#ifdef __cplusplus
+
+#include <cstdint>
+#include <cstring>
+
 namespace twml {
 
-inline int64_t mixDiscreteIdAndValue(int64_t key, int64_t value) {
-  key ^= ((17LL + value) * 2654435761LL);
-  return key;
-}
+    inline int64_t mixDiscreteIdAndValue(int64_t key, int64_t value) {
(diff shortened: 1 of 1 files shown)


nobodyindustries:
How have you tested this? and also tested that it is an actual optimization?

wallbloggerbeing:
> How have you tested this? and also tested that it is an actual optimization?

Yep, went through all the cpp stuff and saw that they were using this header everywhere, even though theres so much ineffecient stuff. Tested it with some sample data and it decreases runtime about 20%. Was a bit frustrated about twitter just wasting resources so I created this pull request.

royeradames:
Great feedback. This is valuable

PR #927, opened 2023-04-01
Switch NSFW text from Python to Notebook format

guibaq:
Changed [NSFW_text.py](https://github.com/twitter/the-algorithm/blob/main/trust_and_safety_models/nsfw/nsfw_text.py) to Jupyter Notebook format.

Indeed, the code reads like it was an extract from a Notebook; it doesn't read naturally in Python format (as it is currently), and it would be more readable to use a Notebook format instead.

In addition:
- added a sample data for the data `df` that seemed to be missing (bug fix). Just filled a few rows to give the idea, I understand that you don't want to share your training data
image

- added some doc to make the file more readable

Could not test the code because it relies on internal Twitter libraries.


spunsosig:
lgtm

PR #936, opened 2023-04-01
jira tickets from comment removed

sandeshtamanq:


VainXploits:
That's great but who asked

src/java/com/twitter/search/common/relevance/entities/GeoObject.java
```diff
@@ -169,11 +169,8 @@ public static boolean approxEquals(GeoObject a, GeoObject b) {
     if (Double.compare(a.radius, b.radius) != 0) {
       return false;
     }
-    if (a.source != b.source) {
-      return false;
-    }
-
-    return true;
+    return a.source != b.source
```
arledesma, line 172: It should be not not, not just not. Not that it matters, but one not is not the same as not not. This is not to be confused with not not not. Alternatively, no not.

arledesma (commented):
To not or not to not, that is the question.

src/java/com/twitter/search/earlybird/document/ThriftDocumentPreprocessor.java
```diff
@@ -69,7 +69,7 @@ public static ThriftDocument preprocess(
           "nullcastFilterSet");
 
   /***
-   * See JIRA SEARCH-7329
+   * 
    */
```
Taz03, line 73: remove the entire comment

src/java/com/twitter/search/earlybird/common/EarlybirdRequestLogger.java
```diff
@@ -55,7 +55,7 @@ protected enum RootOnlyExtraFields {
     DEBUG_MODE,
     CACHE_HIT,
     USER_AGENT,
-    // See JIRA APPSEC-2303 for IP addresses logging
+
```
Taz03, line 58: y an empty new line at the end of enum?

PR #948, opened 2023-04-01
Update README.md

Ahmad-Alkaf:
Fixed misspelled words.


VainXploits:
That's great but who asked

Ahmad-Alkaf:
> That's great but who asked

No one. Small changes do not need to be asked. Take it or leave it 😘
I think this is the point of a Pull Request, innit?

PR #970, opened 2023-04-01
fix: update missing crd in README.md

Saw-YanLinOo:
that why we should thank Elonmusk for this repository.


VainXploits:
That's great but who asked

PR #1076, opened 2023-04-01
fix(tweet): favs ranked range

bozzelliandrea:
Fixed has_gte_10k_favs range with 10k value instead of 1k

### DIFF
--- a/home-mixer/server/src/main/scala/com/twitter/home_mixer/functional_component/decorator/HomeTweetTypePredicates.scala
+++ b/home-mixer/server/src/main/scala/com/twitter/home_mixer/functional_component/decorator/HomeTweetTypePredicates.scala
@@ -160,7 +160,7 @@ object HomeTweetTypePredicates {
     ("has_gte_1k_favs", _.getOrElse(EarlybirdFeature, None).exists(_.favCountV2.exists(_ >= 1000))),
     (
       "has_gte_10k_favs",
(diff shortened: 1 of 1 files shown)


unnimona:
Cool!
Check out https://securityinfinity.com 
Secure your website now

(We are also giving one time opportunity to hack us in 24 hours. Show your skills)
(we initially thought of not writing this to everyone, only to find some keyboard warriors like Kyle Suero giving us boring startup advice. Also this message is dedicated to Twitter team who thinks they can silence everyone in github.com just like they do in Twitter)

home-mixer/server/src/main/scala/com/twitter/home_mixer/functional_component/decorator/HomeTweetTypePredicates.scala
```diff
@@ -160,7 +160,7 @@ object HomeTweetTypePredicates {
     ("has_gte_1k_favs", _.getOrElse(EarlybirdFeature, None).exists(_.favCountV2.exists(_ >= 1000))),
     (
       "has_gte_10k_favs",
-      _.getOrElse(EarlybirdFeature, None).exists(_.favCountV2.exists(_ >= 1000))),
+      _.getOrElse(EarlybirdFeature, None).exists(_.favCountV2.exists(_ >= 10000))),
```
shailrshah, line 163: Nice find! Let's use a numeric literal separator.
`10000` would become `10_000` for more clarity.

PR #1105, opened 2023-04-01
Replaced 1000 with 10000, the real 10k

Dfmaaa:
The keywords on that line suggested that 10000 should be used, not 1000. This commit replaces 1000 with 10000.


sharunkumar:
An actual bug fix no way 🤣

Dfmaaa:
> An actual bug fix no way rofl

felt silly goofy

Dfmaaa:
> An actual bug fix no way rofl
literally saw this bug in a reddit post lmao

Dfmaaa:
> An actual bug fix no way rofl

literally saw this bug in a reddit post lmao

Dfmaaa:
what if no?

PR #1310, opened 2023-04-01
article "THE" added

K2adir:
The name "Social Graph Service" refers to a service that provides access to the social graph, which is a representation of the connections between users in a social network.


K2adir:
The name "Social Graph Service" refers to a service that provides access to the social graph, which is a representation of the connections between users in a social network.

cocomo29:
did u write tests for this? 🤔

K2adir:
> did u write tests for this? 🤔

I asked my mom, she said its ok

cocomo29:
can always count on moms! ✨

PR #1358, opened 2023-04-02
Update NativeUtils.java

Muhanned-Anwar:
Refactor NativeUtils

### DIFF
--- a/ann/src/main/java/com/twitter/ann/faiss/NativeUtils.java
+++ b/ann/src/main/java/com/twitter/ann/faiss/NativeUtils.java
@@ -10,7 +10,7 @@ import java.util.Locale;
 
 public final class NativeUtils {
 
-  private static final int MIN_PREFIX_LENGTH = 3;
+  private static final short MIN_PREFIX_LENGTH = 3;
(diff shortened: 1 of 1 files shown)


ann/src/main/java/com/twitter/ann/faiss/NativeUtils.java
```diff
@@ -19,12 +19,12 @@ private NativeUtils() {
...
 
   private static File unpackLibraryFromJarInternal(String path) throws IOException {
-    if (null == path || !path.startsWith("/")) {
+    if (path == null || !path.startsWith("/")) {
       throw new IllegalArgumentException("The path has to be absolute (start with '/').");
     }
 
     String[] parts = path.split("/");
-    String filename = (parts.length > 1) ? parts[parts.length - 1] : null;
+    String filename = parts.length > 1 ? parts[parts.length - 1] : null;
```
rotexdegba, line 27: Why are you changing **(parts.length > 1)** to **parts.length > 1** ? 

What if there needs to be another condition added in the future like **(parts.length > 1 && someothercondition)**. 

Seems safer to leave the Parenthesis.

PR #1360, opened 2023-04-02
Refactor Function unpackLibraryFromJarInternal in NativeUtils.java

Muhanned-Anwar:
Changes made:

Added a Objects.requireNonNull check for the path parameter to ensure it's not null.
Used Files.createTempFile to create a temporary file instead of manually creating a directory and file. This ensures that the file is created securely and with a unique name.
Removed the unnecessary null check for temporaryDir, as it's now unnecessary with the use of Files.createTempFile.
Simplified the exception handling by catching IOException and NullPointerException separately, and removing the duplicate temp.delete() code.
Used method chaining to simplify the code and reduce nesting.

### DIFF
--- a/ann/src/main/java/com/twitter/ann/faiss/NativeUtils.java
+++ b/ann/src/main/java/com/twitter/ann/faiss/NativeUtils.java
@@ -19,36 +19,33 @@ public final class NativeUtils {
   }
 
   private static File unpackLibraryFromJarInternal(String path) throws IOException {
-    if (null == path || !path.startsWith("/")) {
-      throw new IllegalArgumentException("The path has to be absolute (start with '/').");
+    Objects.requireNonNull(path, "The path cannot be null.");
+    if (!path.startsWith("/")) {
+        throw new IllegalArgumentException("The path has to be absolute (start with '/').");
     }
 
     String[] parts = path.split("/");
-    String filename = (parts.length > 1) ? parts[parts.length - 1] : null;
+    String filename = parts.length > 1 ? parts[parts.length - 1] : null;
 
     if (filename == null || filename.length() < MIN_PREFIX_LENGTH) {
-      throw new IllegalArgumentException("The filename has to be at least 3 characters long.");
+        throw new IllegalArgumentException("The filename has to be at least 3 characters long.");
     }
 
-    if (temporaryDir == null) {
-      temporaryDir = createTempDirectory(NATIVE_FOLDER_PATH_PREFIX);
-      temporaryDir.deleteOnExit();
-    }
-
-    File temp = new File(temporaryDir, filename);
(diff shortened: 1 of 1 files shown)


ann/src/main/java/com/twitter/ann/faiss/NativeUtils.java
```diff
@@ -19,36 +19,33 @@ private NativeUtils() {
   }
 
   private static File unpackLibraryFromJarInternal(String path) throws IOException {
-    if (null == path || !path.startsWith("/")) {
-      throw new IllegalArgumentException("The path has to be absolute (start with '/').");
+    Objects.requireNonNull(path, "The path cannot be null.");
```
shailrshah, line 22: `import java.util.Objects;` is missing. 

https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html

ann/src/main/java/com/twitter/ann/faiss/NativeUtils.java
```diff
@@ -19,36 +19,33 @@ private NativeUtils() {
...
+    if (!path.startsWith("/")) {
+        throw new IllegalArgumentException("The path has to be absolute (start with '/').");
     }
 
     String[] parts = path.split("/");
-    String filename = (parts.length > 1) ? parts[parts.length - 1] : null;
+    String filename = parts.length > 1 ? parts[parts.length - 1] : null;
```
shailrshah, line 28: Let's default to an empty string instead of null. That way we can remove the null check on line 30 and also be sure of no NPEs in the future.

ann/src/main/java/com/twitter/ann/faiss/NativeUtils.java
```diff
@@ -19,36 +19,33 @@ private NativeUtils() {
...
 
     if (filename == null || filename.length() < MIN_PREFIX_LENGTH) {
-      throw new IllegalArgumentException("The filename has to be at least 3 characters long.");
+        throw new IllegalArgumentException("The filename has to be at least 3 characters long.");
     }
```
shailrshah, line 32: We can create a helper method that gets the temp file name from the given path, so that this method remains short.

ann/src/main/java/com/twitter/ann/faiss/NativeUtils.java
```diff
@@ -6,7 +6,7 @@
 import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
-import java.util.Locale;
+import java.util.*;
```
shailrshah, line 9: Would prefer not using wildcard imports.

```
import java.util.Locale;
import java.util.Objects; 
```

https://www.javadude.com/posts/20040522-import-on-demand-is-evil/

shailrshah:
@Muhanned-Anwar, could you give me write permission to your fork? I'd like to make a commit.

PR #1361, opened 2023-04-02
Refactor function loadLibraryFromJar in NativeUtils.java

Muhanned-Anwar:
he main changes are:

Extracted the resource copying logic to a separate method copyResourceToFile.
Renamed the temp variable to tempFile to make its purpose clearer.
Simplified the method by removing unnecessary comments.
Used try-with-resources for the input stream instead of a try-catch block.
Changed the catch block for the NullPointerException to throw a FileNotFoundException instead of catching it and rethrowing.

### DIFF
--- a/ann/src/main/java/com/twitter/ann/faiss/NativeUtils.java
+++ b/ann/src/main/java/com/twitter/ann/faiss/NativeUtils.java
@@ -90,24 +90,26 @@ public final class NativeUtils {
    * @throws FileNotFoundException    If the file could not be found inside the
    *                                  JAR.
    */
-  public static void loadLibraryFromJar(String path) throws IOException {
-    File temp = unpackLibraryFromJarInternal(path);
-
-    try (InputStream is = NativeUtils.class.getResourceAsStream(path)) {
-      Files.copy(is, temp.toPath(), StandardCopyOption.REPLACE_EXISTING);
-    } catch (IOException e) {
-      temp.delete();
-      throw e;
-    } catch (NullPointerException e) {
-      temp.delete();
-      throw new FileNotFoundException("File " + path + " was not found inside JAR.");
+      public static void loadLibraryFromJar(String path) throws IOException {
+        File tempFile = unpackLibraryFromJarInternal(path);
+        try {
+            copyResourceToFile(path, tempFile);
+            System.load(tempFile.getAbsolutePath());
+        } finally {
+            tempFile.deleteOnExit();
+        }
     }
 
-    try {
-      System.load(temp.getAbsolutePath());
-    } finally {
-      temp.deleteOnExit();
-    }
+  private static void copyResourceToFile(String resourcePath, File file) throws IOException {
+      try ( InputStream inputStream = NativeUtils.class.getResourceAsStream(resourcePath)) {
+          Files.copy(inputStream, file.toPath(), StandardCopyOption.REPLACE_EXISTING);
+      } catch (IOException e) {
+          file.delete();
+          throw e;
+      } catch (NullPointerException e) {
+          file.delete();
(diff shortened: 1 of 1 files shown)


HadetTheUndying:
It looks like you're using the GItHub Desktop app. can you please modify your PRs to have more descriptive Titles and your commit messages to be more specific. I don't work for Twitter but from the PR page it looks like you've opened the same PR five times now when that's clearly not the case. But at a glance it looks like you're spamming, when you're clearly not.

Muhanned-Anwar:
> It looks like you're using the GItHub Desktop app. can you please modify your PRs to have more descriptive Titles and your commit messages to be more specific. I don't work for Twitter but from the PR page it looks like you've opened the same PR five times now when that's clearly not the case. But at a glance it looks like you're spamming, when you're clearly not.

A more explanation of the modifications has been added

ann/src/main/java/com/twitter/ann/faiss/NativeUtils.java
```diff
@@ -90,24 +90,26 @@ public static void unpackLibraryFromJar(String path) throws IOException {
...
+  private static void copyResourceToFile(String resourcePath, File file) throws IOException {
+      try ( InputStream inputStream = NativeUtils.class.getResourceAsStream(resourcePath)) {
+          Files.copy(inputStream, file.toPath(), StandardCopyOption.REPLACE_EXISTING);
+      } catch (IOException e) {
+          file.delete();
+          throw e;
+      } catch (NullPointerException e) {
+          file.delete();
+          throw new FileNotFoundException("File " + resourcePath + " was not found inside JAR.");
+      }
```
shailrshah, line 112: How about creating a catch-all block which catches all other exceptions? 

```
catch (Exception e) {
    file.delete();
    throw e;
}
```

ann/src/main/java/com/twitter/ann/faiss/NativeUtils.java
```diff
@@ -90,24 +90,26 @@ public static void unpackLibraryFromJar(String path) throws IOException {
...
-    } catch (IOException e) {
-      temp.delete();
-      throw e;
-    } catch (NullPointerException e) {
-      temp.delete();
-      throw new FileNotFoundException("File " + path + " was not found inside JAR.");
+      public static void loadLibraryFromJar(String path) throws IOException {
+        File tempFile = unpackLibraryFromJarInternal(path);
+        try {
+            copyResourceToFile(path, tempFile);
```
shailrshah, line 96: I like the use of a new method (`copyResourceToFile()`) to abstract the complexities! 

I think the method should be placed outside the try block. We are already handling the deletion of the file if something were to go wrong.

shailrshah: I don't quite understand why we need the logic that's described in `copyResourceToFile()`. `unpackLibraryFromJarInternal()` does the same/similar thing.

uwussimo (commented):
**The updated code appears to be well-written and concise. However, a couple of minor improvements could be made:**

The method name loadLibraryFromJar might not accurately reflect what the method does. The method's name suggests that it loads a library from a JAR, but in reality, it copies a resource from a JAR to a file and then loads the file as a library. Therefore, it might be better to rename the method to something like copyResourceToLibrary to better reflect its functionality.

The IOException thrown in the finally block when deleting the temporary file could mask any exceptions thrown during the System.load call. To avoid this, it might be better to catch and handle any exceptions thrown during the System.load call explicitly, rather than relying on the finally block.

These are minor suggestions, and the updated code is already well-written and efficient. Therefore, they are not critical, and the code is recommended to be approved as is.

PR #1370, opened 2023-04-02
too many featureName variables

lgon070:
there just too many lets get rid of them

### DIFF
--- a/src/java/com/twitter/search/common/util/ml/StringMapBasedLinearModel.java
+++ b/src/java/com/twitter/search/common/util/ml/StringMapBasedLinearModel.java
@@ -65,13 +65,12 @@ public class StringMapBasedLinearModel implements MapBasedLinearModel<String> {
   public float score(Map<String, Float> values) {
     float score = 0.0f;
     for (Map.Entry<String, Float> value : values.entrySet()) {
-      String featureName = value.getKey();
-      float weight = getWeight(featureName);
+      float weight = getWeight(value.getKey());
       if (weight != 0.0f) {
         score += weight * value.getValue();
         if (LOG.isDebugEnabled()) {
           LOG.debug(String.format("%s = %.3f * %.3f = %.3f, ",
-              featureName, weight, value.getValue(),
+          value.getKey(), weight, value.getValue(),
               weight * value.getValue()));
         }
       }


src/java/com/twitter/search/common/util/ml/StringMapBasedLinearModel.java
```diff
@@ -65,13 +65,12 @@ protected Map<String, Float> getWeights() {
   public float score(Map<String, Float> values) {
```
shailrshah, line 65: Let's rename `values` to `feature_to_values`. 

That way, the for loop becomes:
```
for (Map.Entry<String, Float> entry : featureToValues.entrySet()) {
    String feature = entry.getKey();
    Float value = entry.getValue();
    Float weight = getWeight(feature);
    ...
}
```

src/java/com/twitter/search/common/util/ml/StringMapBasedLinearModel.java
```diff
@@ -65,13 +65,12 @@ protected Map<String, Float> getWeights() {
...
     float score = 0.0f;
     for (Map.Entry<String, Float> value : values.entrySet()) {
-      String featureName = value.getKey();
```
rotexdegba, line 68: The featureName variable makes the code more understandable.

**value.getKey()** is not descriptive by itself.

Seems better to keep the featureName variable

src/java/com/twitter/search/common/util/ml/StringMapBasedLinearModel.java
```diff
@@ -65,13 +65,12 @@ protected Map<String, Float> getWeights() {
...
-      float weight = getWeight(featureName);
+      float weight = getWeight(value.getKey());
```
shailrshah, line 68: This is confusing because `value` is actually an `Entry` and not the value of a `Map` or `Entry`.

src/java/com/twitter/search/common/util/ml/StringMapBasedLinearModel.java
lgon070: You're not wrong, but would it not be funny to confuse future people looking at this. Ripe opportunity for it.

PR #1429, opened 2023-04-02
Update split files into more files

ImperiumTakp:


uwussimo (commented):
Given the significant size of the commit, it would be challenging to ensure that all the changes have been reviewed correctly, and any potential issues have been identified and addressed. Therefore, I would suggest splitting the changes into smaller, more manageable commits that can be reviewed more effectively.

palttamas:
Not enough separation. For better maintainability, can we get each line onto its own file please?

PR #1440, opened 2023-04-02
possible polynomial runtime due to backtracking

elaneri:
Hi, I found this could be a issue

Make sure the regex used here, which is vulnerable to polynomial runtime due to backtracking, cannot lead to denial of service.

### DIFF
--- a/src/java/com/twitter/search/feature_update_service/FeatureUpdateController.java
+++ b/src/java/com/twitter/search/feature_update_service/FeatureUpdateController.java
@@ -200,10 +200,10 @@ public class FeatureUpdateController implements FeatureUpdateService.ServiceIfac
   }
 
   private List<PenguinVersion> getPenguinVersions(String penguinVersionsStr) {
-    String[] tokens = penguinVersionsStr.split("\\s*,\\s*");
+    String[] tokens = penguinVersionsStr.split(",");
     List<PenguinVersion> listOfPenguinVersions = Lists.newArrayListWithCapacity(tokens.length);
     for (String token : tokens) {
-      listOfPenguinVersions.add(PenguinVersion.valueOf(token.toUpperCase()));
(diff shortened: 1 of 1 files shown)


ElijahPepe:
The general idea here is lost, because the time complexity is equivalent for both implementations. However, I will add that Java uses a [lightweight implementation](https://hg.openjdk.org/jdk7/jdk7/jdk/rev/1ff977b938e5) of `.split()` for single-character strings (that are not meta characters) that doesn't use RegEx, to my knowledge. In this case, trimming the string afterwards would be faster. LGTM.

ElijahPepe:
The general idea here is lost, because the time complexity is equivalent for both implementations. However, I will add that Java uses a [lightweight implementation](https://hg.openjdk.org/jdk7/jdk7/jdk/rev/1ff977b938e5) of `.split()` for single-character strings (that are not meta characters) that doesn't use RegEx, to my knowledge. In this case, trimming the string afterwards would be faster. LGTM.

uwussimo:
Based on the potential issue described, the update to the code is a good improvement and should be approved.

Adding the `trim()` method call on the token ensures that any leading or trailing whitespaces in the input string are removed before converting the string to an enum value. This helps avoid any possible IllegalArgumentException that could have occurred if there were any leading or trailing whitespaces in the input string.

The update also doesn't change the overall behavior of the method and doesn't introduce any new issues or side effects. Therefore, it is recommended to approve the update to ensure the code is more robust and less prone to errors.

ElijahPepe:
@uwussimo Not sure where you're getting some of that from, this is a function taking a string and splitting it by whitespace padded commas.

elaneri:
Hi , @ElijahPepe, you can find more information about this https://rules.sonarsource.com/java/RSPEC-5852
regards

ElijahPepe:
This code is used in a specific, uncontrollable string that wouldn't be susceptible to a ReDoS. Not saying that the pull request isn't valid, but there's no security risk here.

PR #1452, opened 2023-04-02
Refactored code to improve performance and readability.

wgw0:
## Refactored code to improve performance and readability.

Refactored the `getGeohashAndCountryCode` method in the `UserLocationFetcher` class to improve performance and readability. Specifically, made the following changes:

- Moved the creation of the stats counters (`totalRequestsCounter`, `emptyResponsesCounter`, `locationServiceExceptionCounter`, `reverseGeocodeExceptionCounter`) to where they are used, inside the `getGeohashAndCountryCode` method, to avoid creating them unnecessarily.

- Simplified the creation of `lscLocationStitch` and `ipLocationStitch` by using `Stitch.collect` instead of wrapping the `Option` in a `Stitch` and using `Stitch.collect` inside it.

- Combined the matching of `geohash` and `countryCode` into a single pattern match on a tuple, instead of using `flatMap` and `orElse`.

- Removed the `private val` definitions for the stats counters and used them directly in the method, to avoid unnecessary indirection.

These changes improve the performance of the `getGeohashAndCountryCode` method by reducing unnecessary object creation and function calls, while also making the code easier to read and maintain.

### DIFF
--- a/follow-recommendations-service/common/src/main/scala/com/twitter/follow_recommendations/common/clients/geoduck/UserLocationFetcher.scala
+++ b/follow-recommendations-service/common/src/main/scala/com/twitter/follow_recommendations/common/clients/geoduck/UserLocationFetcher.scala
@@ -13,10 +13,6 @@ class UserLocationFetcher @Inject() (
   statsReceiver: StatsReceiver) {
 
   private val stats = statsReceiver.scope("user_location_fetcher")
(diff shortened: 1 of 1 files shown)


MurphyAdam:
This does not look like a shipost, therefore Doesn't LGT

wgw0:
> This does not look like a shipost, therefore Doesn't LGT

Not sure what LGT means.

PR #1491, opened 2023-04-03
fixed capitlization in a file

Minejerik:


joao-pedro-alves:
You should fix your huge amount of free time...

hoangphuc05:
I think we should lowercase all instances of tweet and retweet instead

PR #1493, opened 2023-04-03
Elongated

Cillian1234:
Changed every instance of 'e' to "elon"


NsdHSO:
I see that you added the `.idea` directory, why?
@twiter-team Please add `.idea` to `.gitignore`

huseinnashr:
LGTM

awwong1:
some high impact code right here, better print it out double sided

rickvian:
you gotta follow the convention, it should be camelCase

please put Elon if there are any precedding text e.g thElon

iswarmondal:
Respect👮! Take that AI

moonrailgun:
I have no idea why xxxe will be `xxxelon`.

I has some problem, i think. because its will be loop themself. like `xxxelonlonlonlonlonlon......`

ibreathebsb:
> I have no idea why xxxe will be `xxxelon`.
> 
> I has some problem, i think. because its will be loop themself. like `xxxelonlonlonlonlonlon......`
> 
> BTW, another problem is i cannot review this pr, i think microsoft programmer should test their page in benchmark image

works like "nnoremap" in vim

xiaoyu2006:
LGTM. Nice work dude!

PR #1503, opened 2023-04-03
Fix a few typos

bnzone:
Just fixing a few typos ✌🏻


Neptuneia:
the only thing you need to fix is your free time

bnzone:
> the only thing you need to fix is your free time

Oh, thanks to a good extension it only took about a minute 😉

Svehla:
ty seš teda ale hrozná piča....

PR #1521, opened 2023-04-03
Refactored code to improve performance and readability - LocationServiceClient.scala

wgw0:
## Improvements Made in the Code

The following changes have been made to improve the performance and readability of the Scala code:

- The `Stitch` calls in the `getGeohashAndCountryCode` method have been simplified using a `for` comprehension, which is more readable.
- The `getGeohashFromTransactionLocation` function now returns a `Stitch` instead of an `Option`, which is more consistent with the return type of `getGeohashAndCountryCode`.
- The code has been formatted to improve readability, using consistent indentation and spacing.
- The `flatMap` calls in `getGeohashFromTransactionLocation` have been replaced with pattern matching, which is more concise and readable.
- Unused variables and imports have been removed.


uwussimo:
The updated code is a good improvement. Here are some specific suggestions:

1. Using `for`-comprehension instead of `map-flatMap` chaining is a good way to make the code more readable and easier to follow.
2. Using `Stitch.from` is a better way to wrap the `Future` into a `Stitch` monad. This way, we can avoid using the Stitch.callFuture method.
3. Using pattern matching instead of `flatMap` and `map` can make the code more concise and easier to read. In particular, the `getGeohashFromTransactionLocation` method can be simplified.
4. The use of `Stitch.value` to create a new `Stitch` value is a good practice that makes the code more consistent.

> Overall, the updated code is a good improvement over the original code.

PR #1523, opened 2023-04-03
Optimized mapper.rs

ismoilovdevml:
The code has been optimized. Now it's faster and more efficient

### DIFF
--- a/navi/segdense/src/mapper.rs
+++ b/navi/segdense/src/mapper.rs
@@ -19,13 +19,19 @@ pub struct FeatureMapper {
 impl FeatureMapper {
     pub fn new() -> FeatureMapper {
         FeatureMapper {
-            map: HashMap::new()
+            map: HashMap::new(),
+        }
+    }
+
+    pub fn with_capacity(capacity: usize) -> FeatureMapper {
+        FeatureMapper {
+            map: HashMap::with_capacity(capacity),
         }
     }
 }
 
 pub trait MapWriter {
-    fn set(&mut self, feature_id: i64, info: FeatureInfo); 
+    fn set(&mut self, feature_id: i64, info: FeatureInfo);
 }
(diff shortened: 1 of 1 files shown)


wgw0 (commented):
## Refactored your code for performance

- Importing the `HashMap` module to avoid namespace conflicts and make the code more readable.
- Replacing `pub fn set`'s `entry` method with `insert` method since we don't need to check if the key already exists.
- Reordering the traits `MapReader` and `MapWriter` to group them together.
- Renaming `FeatureMapper` methods to `new` and `with_capacity` to match Rust conventions.
- Adding `Self` as the return type for `new` and `with_capacity` to avoid redundant code.
- Adding `&self` as the parameter type for `get` to indicate that it doesn't modify the `FeatureMapper` object.


```rust
use std::collections::HashMap;

pub struct FeatureMapper {
    map: HashMap<i64, FeatureInfo>,
}

impl FeatureMapper {
    pub fn new() -> Self {
        Self {
            map: HashMap::new(),
        }
    }

    pub fn with_capacity(capacity: usize) -> Self {
        Self {
            map: HashMap::with_capacity(capacity),
        }
    }
}

pub trait MapWriter {
    fn set(&mut self, feature_id: i64, info: FeatureInfo);
}

pub trait MapReader {
    fn get(&self, feature_id: i64) -> Option<&FeatureInfo>;
}

impl MapReader for FeatureMapper {
    fn get(&self, feature_id: i64) -> Option<&FeatureInfo> {
        self.map.get(&feature_id)
    }
}

impl MapWriter for FeatureMapper {
    fn set(&mut self, feature_id: i64, info: FeatureInfo) {
        self.map.insert(feature_id, info);
    }
}
```

0ex-d (commented):
😂

PR #1535, opened 2023-04-03
Constant Renamed For Better Understanding

iahmadhabibx:


cynical-incendiary:
There are 98 syntax errors and 14 typos in your changes; please rewrite this project from scratch and force-push onto the relevant branch so we can merge this. Afterwards, you will receive the free blue tick as a verified community contributor. Thanks.

Sincerely,
Elon "Twitter" Musk

jtbx:
This just makes the code look worse and less consistent

DanielJSottile:
> This just makes the code look worse and less consistent

In all seriousness, I agree, this doesn't necessarily add anything to the codebase.  I would even say that this makes it harder to understand exactly what the variable does.

src/python/twitter/deepbird/projects/timelines/configs/recap_earlybird/feature_config.py
```diff
@@ -65,8 +65,8 @@ def get_feature_config(data_spec_path, label):
   ).add_labels([
     label,                                   # Tensor index: 0
     "recap.engagement.is_clicked",           # Tensor index: 1
-    "recap.engagement.is_favorited",         # Tensor index: 2
-    "recap.engagement.is_open_linked",       # Tensor index: 3
+    "recap.engagement.is_favorite",         # Tensor index: 2
```
nmummau, line 68: You probably didn't notice, but this change impacted the whitespace which aligned the comments.

PR #1552, opened 2023-04-03
4% image compression on system-diagram.png

ZeroDayDaemon:
This PR addresses the issue of a binary image file being uncompressed within the repository, and rectifies it by applying a compression of 4% to the relevant binary image file.


ilovelinux:
This PR will allow us to save a few kb with the drawback to keep the older image in the git history, actually "doubling" the image size in the repository. IMHO it's not worth it.

ZeroDayDaemon:
> This PR will allow us to save a few kb with the drawback to keep the older image in the git history, actually "doubling" the image size in the repository. IMHO it's not worth it.

git has deltas for binary files, so, it would not be nearly double the size within the repository including commit history.

ilovelinux:
> > This PR will allow us to save a few kb with the drawback to keep the older image in the git history, actually "doubling" the image size in the repository. IMHO it's not worth it.
> 
> git has deltas for binary files, so, it would not be nearly double the size within the repository including commit history.

It doesn't work with images and compression, but I might be wrong.

ZeroDayDaemon:
> It doesn't work with images and compression, but I might be wrong.

It possibly depends on how compression is done, I do believe git handles all binary files the same notwithstanding settings being changed deliberately for specific binary file types.

Regardless, I believe the compression here is necessary, assuming I am incorrect and it does make the repository larger, someone not interested in the repository history would still receive the benefits, and assuming the image is included in production environments, for what reason I wouldn't know, perhaps it makes the computers run faster, It would also receive the benefits. 

Overall, the improvement to the binary file is a positive change for the file itself, regardless of git history, and has several cases of improving certain aspects of the codebase.

ilovelinux:
@ZeroDayDaemon I edited my first comment as soon as I noticed that I misunderstood that 4%. I thought it was how much it shrank but it is the compressed file size compared to the original one.

Worth it, looks good to me.

Thanks!

PR #1601, opened 2023-04-03
Fixes in comments

adityaroshandash:
1. Removed spaces in comments to look appropriate.


ilovelinux:
This PR should be split into two different PRs:
1) Remove empty comment's newlines.
2) Fix "neigbors" typo

I disagree with removing empty comment's newlines since they separate the function description from the description of the arguments and the return value, of the function.

neigbors fix looks good to me.

PR #1604, opened 2023-04-04
Add a most used emoji to predict nsfw content 🍆

DamianRz:
I feel that this emoji in this situation would be helpful

### DIFF
--- a/trust_and_safety_models/nsfw/nsfw_text.py
+++ b/trust_and_safety_models/nsfw/nsfw_text.py
@@ -134,7 +134,7 @@ history = model.fit(
   steps_per_epoch=len(ds_train)
 )
 
-model.predict(["xxx 🍑"])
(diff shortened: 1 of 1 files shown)


islemci:
what do i do if i want to show off my homegrown eggplant... 😭😭😭

src-ry:
> what do i do if i want to show off my homegrown eggplant... 😭😭😭

he don't care

AleAtSoftplace:
At this point one could also use 💦

MatthieuHernandez:
This pull request does not change the training dataset so it does not change the prediction model.
Just changing a test does almost nothing, this PR is useless and should be closed.

delgh1 (commented):
More can be added, probably should change

src-ry (commented):
man add more stuff pls

PR #1606, opened 2023-04-04
Typo fixes

srahman5:
- "simlarity" -> "similarity"
- Whitespace linted


DanielJefferson1990:
Maybe you should merge with utsavDave97:typo-fixes

ilovelinux:
@DanielJefferson1990 I think it's fine as it is since they are two different authors working on the same issue but on two different parts of code.

LGMT

PR #1612, opened 2023-04-04
Update README.md

Jinyi-Liu:
In the code, the negative labels are all removed. The one with label 0 should be non-positive.

### DIFF
--- a/src/scala/com/twitter/interaction_graph/README.md
+++ b/src/scala/com/twitter/interaction_graph/README.md
@@ -4,7 +4,7 @@ This project builds a machine learning model using a gradient boosting tree clas
 
 The algorithm works by first creating a labeled dataset of user interactions from a graph of Twitter users. This graph is represented in a BigQuery table where each row represents a directed edge between two users, along with various features such as the number of tweets, follows, favorites, and other metrics related to user behavior.
 
-To create the labeled dataset, the algorithm first selects a set of candidate interactions by identifying all edges that were active during a certain time period. It then joins this candidate set with a set of labeled interactions that occurred one day after the candidate period. Positive interactions are labeled as "1" and negative interactions are labeled as "0". The resulting labeled dataset is then used to train a boosted tree classifier model.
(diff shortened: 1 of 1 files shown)


ilovelinux:
"non-positives" includes 0, "negatives" doesn't. This means that "0" would be labeled as "1" since it's not negative.

Could you check that this is the actual behavior?

Jinyi-Liu:
> "non-positives" includes 0, "negatives" doesn't. This means that "0" would be labeled as "1" since it's not negative.
> 
> Could you check that this is the actual behavior?

Here is the link for the code.
https://github.com/twitter/the-algorithm/blob/d1cab28a1044a147a107ae067890850041956777/src/scala/com/twitter/interaction_graph/bqe/training/labeled_candidates.sql

SELECT 
  F.source_id,
  F.destination_id,
  CASE WHEN P.source_id IS NULL THEN 0 ELSE 1 END AS label,
  num_days,
  num_tweets,
  num_follows,
  num_favorites,
  num_tweet_clicks,
  num_profile_views,
  days_since_last_interaction,
  label_types,
  date_candidates AS ds
FROM `twttr-recos-ml-prod.realgraph.candidates_sampled` F
LEFT JOIN label_positive P USING(source_id, destination_id)
LEFT JOIN label_negative N USING(source_id, destination_id)
WHERE N.source_id IS NULL AND N.destination_id IS NULL
AND F.ds=date_candidates
;

At line 39, CASE WHEN P.source_id IS NULL THEN 0 ELSE 1 END AS label,
So if the row chosen from ``twttr-recos-ml-prod.realgraph.candidates_sampled'' has no positive label then it's labeled with 0. 

At line 52, WHERE N.source_id IS NULL AND N.destination_id IS NULL will ignore all the rows from  ``twttr-recos-ml-prod.realgraph.candidates_sampled'' with negative label.

So I think there are originally three labels {positive, neutral, negative}. The code here ignores all the rows with negative labels.

Issue #1630, opened 2023-04-04
Change method to static method


Ibrahim2750mi:
Currently [this](https://github.com/twitter/the-algorithm/blob/main/src/python/twitter/deepbird/projects/timelines/scripts/models/earlybird/lolly/parsers.py#L127) method returns just a regex which doesn't uses any class attribute. According to pythonic code this should be a `staticmethod`.

wallbloggerbeing:
> 

yeah ml in rust

SqrtTan:
> Python sucks, rewrite it in RUST!

Why not use COBOL?

sokorototo:
COBOL = Common Bullshit Oriented Language

PR #1636, opened 2023-04-04
Change method to static method #1630

csankhala:
Fix is provided for reported issue: https://github.com/twitter/the-algorithm/issues/1630


kapilepatel:
Verified, looks good to me @twitter-service  please merge

PR #1646, opened 2023-04-04
Removed dead code in navi

jamesjoegraham:


dzhao:
TensorInputEnum is needed.

PR #1657, opened 2023-04-04
minor improvement: EarlybirdWireModule.java readability

MatheusGarvao:
improve readability

### DIFF
--- a/src/java/com/twitter/search/earlybird/factory/EarlybirdWireModule.java
+++ b/src/java/com/twitter/search/earlybird/factory/EarlybirdWireModule.java
@@ -422,8 +422,8 @@ public class EarlybirdWireModule {
   public ScoringModelsManager provideScoringModelsManager(
       SearchStatsReceiver serverStats,
       EarlybirdIndexConfig earlybirdIndexConfig) {
-    boolean modelsEnabled = EarlybirdConfig.getBool("scoring_models_enabled", false);
-    if (!modelsEnabled) {
+    boolean modelsDisabled = !EarlybirdConfig.getBool("scoring_models_enabled", false);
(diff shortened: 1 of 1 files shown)


src/java/com/twitter/search/earlybird/factory/EarlybirdWireModule.java
```diff
@@ -422,8 +422,8 @@ public EarlybirdDarkProxy provideEarlybirdDarkProxy(
   public ScoringModelsManager provideScoringModelsManager(
       SearchStatsReceiver serverStats,
       EarlybirdIndexConfig earlybirdIndexConfig) {
-    boolean modelsEnabled = EarlybirdConfig.getBool("scoring_models_enabled", false);
-    if (!modelsEnabled) {
+    boolean modelsDisabled = !EarlybirdConfig.getBool("scoring_models_enabled", false);
```
shailrshah, line 425: Nit: Rename boolean to `areModelsDisabled` to clearly indicate a boolean variable.

Same for line 453

ProKashif (commented):
LGTM

src/java/com/twitter/search/earlybird/factory/EarlybirdWireModule.java
adrianarandac: The use of the negation operator (!) on the value returned by EarlybirdConfig.getBool() means that the modelsDisabled variable will be assigned true if the value of "scoring_models_enabled" is false. This can be confusing, as the name of the variable suggests that it would be false if the models are disabled. It might be more clear to assign the value returned by EarlybirdConfig.getBool() directly to modelsDisabled, without using the negation operator, and then adjust the variable name accordingly.

PR #1524, opened 2023-04-03
Update SpaceSafetyLabelType.scala

danielhaim1:
Summary
=======

This content provides an overview of the Visibility Filtering algorithm and its types, which represent different characteristics of content that may pose a risk or harm to users to support legal compliance, improve product quality, and increase user trust through different remediations.

The algorithm utilizes three primary concepts: `SafetyLevel` `SafetyLabel` and `SafetyLabelType` It also defines several safety label types that can be grouped into three categories: _Content Classification_, _Content Moderation_, and _Status_. 

This article further explains the `UkraineCrisisTopic` safety label type, and its removal from the list of defined categories, the language used in the code which is not politically correct, and addresses recent staff cuts in Twitter's Trust and Safety and Global Content Moderation team which may hinder the review and action of flagged tweets.

Update (April 6th, 2023)
==================

As of April 5th, 2023, the changes discussed in this article [have been merged](https://github.com/twitter/the-algorithm/commit/3f6974687f0c16923da7e27b2c467abd0a109a1c) into the Visibility Filtering algorithm. The safety label type `UkraineCrisisTopic` has been removed from the code, and the categorization of safety label types has been updated to ensure consistency and accuracy.

I would like to thank everyone for their contributions and effort in getting this pull request merged, and I hope these changes will help improve the safety and trust of Twitter's platform for all users. Your effort and dedication have been instrumental in making these changes possible, and I am honored to be a part of such a supportive and collaborative community.

Visibility Filtering
=============

Visibility Filtering is a rule engine that centrally instructs the client on adjusting the display of specific content at read time. This library is responsible for _"filtering content to support legal compliance, improve product quality, increase user trust, protect revenue through hard-filtering, visible product treatments, and coarse-grained downranking"_.

### Types

The library utilizes three primary concepts to support this:

* `SafetyLevel`: represents the context in which content is viewed
* `SafetyLabel`: identifies content as potentially harmful
* `SafetyLabelType` specifies the type of policy violation
* `SpaceSafetyLabelType` groups safety labels based on content characteristics

### Categories

It is possible to logically categorize them into three groups:

1. **Content Classification**
   * `CoordinatedHarmfulActivityHighRecall`
   * `HatefulHighRecall`
   * `ViolenceHighRecall`
   * `CivicIntegrityMisinfo`
   * `MedicalMisinfo`
   * `GenericMisinfo`
   * `MisleadingHighRecall`
   * `NsfwHighPrecision`
   * `NsfwHighRecall`
[...]

PR #1659, opened 2023-04-04
Renames UkraineCrisisTopic safety label to RussianInvasionOfUkraineTopic

mptorz:
More context about the renamed safety label and the ongoing discussion about possible removal of it can be tracked in https://github.com/twitter/the-algorithm/pull/1524. Until decision is reached there I suggest to at least name it appropriately.

This safety label refers to the ongoing [Russian invasion of Ukraine](https://en.wikipedia.org/wiki/Russian_invasion_of_Ukraine) and should be named as such. Current naming is vague and easy to misinterpret. `RussianInvasionOfUkraineTopic` is a more accurate name.


paxperscientiam:
OP is 100% correct. Please re-open.

Nightcaat:
It was closed in https://github.com/twitter/the-algorithm/commit/3f6974687f0c16923da7e27b2c467abd0a109a1c because the label is [deprecated](https://github.com/twitter/the-algorithm/commit/3f6974687f0c16923da7e27b2c467abd0a109a1c#diff-9d224a3fac44a2f4379b3a970c7ef98f716170a3193408ac50a15805db848793R39), and as such, there is no longer a need for a more accurate name.

PR #1669, opened 2023-04-05
refine with static list

ChiaChinHo:
1. refine with private static variable, instead of compare value one by one


ProKashif (commented):
LGTM

guitar-etc:
Unless the list is used elsewhere, It seems more performance to keep the code as is.
In order to reduce duplication, it would be possible to have the latter function use the prior function.

PR #1693, opened 2023-04-06
Replaced the source code with Super Mario 64

RatserX:
The Super Mario Bros. Movie just released, and the algorithm is using the same old code? This is absurd.
Here, take the Super Mario 64 source code instead and go celebrate!


wiichicken:
legend for **actually** replacing it with the real source code

LiamChat:
I believe that this will change twitter for the better

uzbeki:
a little funny, but lameeeeeeee

mylesbartlett72:
> I believe that this will change twitter for the better

yeah this would finally kill the birdsite when Nintendo DMCAs them into nonexistence

PR #1696, opened 2023-04-06
Refactor AlgorithmTypeConstants to use AlgorithmType enums and improve readability

BankkRoll:
- Created type aliases `AlgorithmTypeSet` and `AlgorithmToTypeMap` for better readability
- Removed the unnecessary `.toString()` method calls in the map keys
- Changed the type of the map keys from `String` to `Algorithm` enum values
- Changed the value type of the map from `Set[AlgorithmType.Value]` to just `AlgorithmTypeSet`
- Removed the comments indicating the categories of the algorithms, as they were already indicated in the `AlgorithmType` enum
- Reordered the algorithms alphabetically by their feedback token
- Simplified the `getAlgorithmTypes` method to remove the map call and use a default value instead of returning an empty set
- Removed maybe a type line comment `//` in line 67 "AlgorithmToFeedbackTokenMap(PopGeohashRealGraph).toString -> Set(AlgorithmType.Geo)," was commented out


BankkRoll:
stale

PR #1698, opened 2023-04-06
Refactor UserStateClient: Improved Readability, Modularity, and Scalability

BankkRoll:
- Added type annotations to variable declarations
- Used pattern matching instead of an if-else statement in `getUserState`
- Added type annotations to the return values of methods and variables
- Removed a redundant `toString` call
- Used `Stitch(None)` instead of `Stitch(Some(None))` in the `.rescue` block


BankkRoll:
stale

PR #1725, opened 2023-04-06
Tiny fix the error handling in Rust build file

0ex-d:
We can use `anyhow::Error` over `Box<dyn std::error::Error>` which can support more error handling, and since we already use `anyhow` crate for handling `Result` trait in other files.


harshsingh32 (commented):
looks fine

PR #1729, opened 2023-04-07
Fix Error Message 

pierizvi:
With this change, the error message will be more clear and accurate, making it easier to understand and fix any issues that may arise from incorrect metric values.

### DIFF
--- a/ann/src/main/python/dataflow/faiss_index_bq_dataset.py
+++ b/ann/src/main/python/dataflow/faiss_index_bq_dataset.py
@@ -91,7 +91,7 @@ def parse_metric(config):
   elif metric_str == "linf":
     return faiss.METRIC_Linf
   else:
-    raise Exception(f"Uknown metric: {metric_str}")
+    raise Exception(f"Unknown metric: {metric_str}")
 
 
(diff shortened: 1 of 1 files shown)


GTF0:
> With this change, the error message will be more clear and accurate, making it easier to understand and fix any issues that may arise from incorrect metric values.

You're incredibly talented.

ann/src/main/python/dataflow/faiss_index_bq_dataset.py
```diff
@@ -91,7 +91,7 @@ def parse_metric(config):
   elif metric_str == "linf":
     return faiss.METRIC_Linf
   else:
-    raise Exception(f"Uknown metric: {metric_str}")
```
MolarFox, line 94: Actually I believe this is an intentional reference to [Uknown Reggae Band](https://en.wikipedia.org/wiki/File:Uknown_reggae_band_live.jpg), likely added here that anyone who is forced to read or interact with this code or its logs is also forced to think about Reggae for a little while

ChanderJindal (commented):
I wanna know if this type of commits are accepted during Hactoberfest.

PR #1734, opened 2023-04-07
Avoid raw get

einnocent:


gran4 (commented):
The code is better with the match instead of the if statements. Looks good to me.

pierizvi:
LGTM

Hokutosei:
LGTM

Issue #1742, opened 2023-04-08
Add tag for War Misinformation


NicholasFlorian:
**Is your feature request related to a problem? Please describe.**
It would be nice to create a unique tag separate from civic misinformation to help document war misinformation. War misinformation is quite common on the internet and unverified information about conflicts should be tagged correctly.

**Describe the solution you'd like**
A new tag should be created for war misinformation

**Describe alternatives you've considered**

**Additional context**
I know that the Ukraine tag was deprecated in this [PR](https://github.com/twitter/the-algorithm/issues/1489) so the tag should be removed premaritally in refactoring. A proper tag should be added for all war misinformation.

TolstoyDotCom:
Scroll through the tweets by yoyoel. He's (thankfully) not at Twitter any more, but people like him are still there. Do you really think pro-censorship airheads like him should be in charge of deciding what is and what isn't war "misinfo"? What about the Ukrainian, USA, or Russian militaries?

The proper way to deal with "misinfo" is to show it wrong. Twitter - by design - makes that difficult to do because they favor those with power while silencing those without power.

nobodyindustries:
OK bots

ValZapod:
>from civic misinformation to help document war misinformation.

Both USA and Russia are lying in their proxy conflict. Even the SI GAMMA classified sources suggest there is 8k difference is lower and higher estimates of Russians https://img.4plebs.org/boards/pol/image/1680/70/1680708793186116.png

vwellum:
@ValZapod Please stop. GitHub is not a place for Russian propaganda.
Government should really release these documents because so many tampered versions circle around the internet that it's hard to catch up to some people.

ValZapod:
>GitHub is not a place for Russian propaganda.

You mean 4chan propaganda. It was a joke on 4chan

![1680848254522240](https://user-images.githubusercontent.com/31514790/232466199-9d1f3d99-aa51-4355-b407-38e8924042dd.png)

Lucienne1984:
There is a fundamental flaw with this plan. Nobody, not even the high commands of Ukraine or Russia, has exact information on what's going on. Everything you see is an approximation. Nobody, save for god if he exists, has the ability to give accurate info on the war, therefore nobody can fact-check properly to do this either. Obvious edits like what Valzapod posted can be spotted, but we don't even know if the unedited document is fully correct.

sylentpunk:
[...]

Issue #1489, opened 2023-04-02
Safety Label Types and Categorization in Code


danielhaim1:
Please reference to #1524

danielhaim1:
Slava Ukraini 🇺🇦

webknjaz:
@danielhaim1 thanks for the analysis!

webknjaz:
> the Ukraine crisis

I must also add that this is a poor word choice. People who want to represent the objective reality or support Ukraine, aren't afraid to call it invasion/war, as it is. The word "crisis" is mostly used by the ruscist supporters or "neutral" media (which usually also counts as working against Ukraine) — would one call an armed robbery of a convenience store a "store crisis"? Why not?

danielhaim1:
@webknjaz while considering the safety concerns and the fact that the file is easy to read and stands out, I found myself wondering if the code was added around 2014. Regardless of politics, it is clear that the code does not belong in this file. Additionally, I believe this file is frequently accessed.

> > the Ukraine crisis
> 
> I must also add that this is a poor word choice. People who want to represent the objective reality or support Ukraine, aren't afraid to call it invasion/war, as it is. The word "crisis" is mostly used by the ruscist supporters or "neutral" media (which usually also counts as working against Ukraine) — would one call an armed robbery of a convenience store a "store crisis"? Why not?

@webknjaz while considering the safety concerns and the fact that the file is easy to read and stands out, I found myself wondering if the code was added around 2014. Regardless of politics, it is clear that the code does not belong in this file. Additionally, I believe this file is frequently accessed.

danielhaim1:
> > the Ukraine crisis
> 
> I must also add that this is a poor word choice. People who want to represent the objective reality or support Ukraine, aren't afraid to call it invasion/war, as it is. The word "crisis" is mostly used by the ruscist supporters or "neutral" media (which usually also counts as working against Ukraine) — would one call an armed robbery of a convenience store a "store crisis"? Why not?

while considering the safety concerns and the fact that the file is easy to read and stands out, I found myself wondering if the code was added around 2014. Regardless of politics, it is clear that the code does not belong in this file. Additionally, I believe this file is frequently accessed.

webknjaz:
> I found myself wondering if the code was added around 2014

Well, it was still a war in 2014.

danielhaim1:
Why did you close the ticket @maksimru ?

chrisidakwo:
[...]

PR #1743, opened 2023-04-08
Add War Misinfo

NicholasFlorian:
It would be nice to create a unique tag separate from civic misinformation to help document war misinformation. War misinformation is quite common on the internet and unverified information about conflicts should be tagged correctly.

I know that the Ukraine tag was deprecated in this https://github.com/twitter/the-algorithm/issues/1489 so the tag should be removed permanently in refactoring. A proper tag should be added for all war misinformation.

See my issue [here](https://github.com/twitter/the-algorithm/issues/1742).


caleb-depotanalytics:
Why is this at the cost of the "Crisis" misinformation @NicholasFlorian ? It seems like a Hurricane and an Invasion should have two separate tags.

harshsingh32 (commented):
Looks fine

harshsingh32 (commented):
reviewed

PR #1753, opened 2023-04-10
Navi clippy fixes

ZutrixPog:
- fixed formatting and clippy problems
- cleaned the code a bit


dzhao:
can you update the pr with the Cargo.lock removed? We prefer to not include that in the source. Thanks!

ZutrixPog:
just removed cargo.lock

PR #1768, opened 2023-04-11
* [COMMENT] - ShadowBan shoud be used to ban bad users, not to become a tool for online criminal militias silence opponents

richardomus:
There is clearly a PROBLEM with this approach: ShadowBan is supposed to ban abusers from the platform, but with this code, it can be used by malicious groups and virtual militias to produce a coordinated ATTACK to silence someone they don't like. There are several reports of this on Twitter, people who defend human rights or produce anti-racist content are being attacked by Follow -> unfollow or block / mute attacks, producing the unbalance equation needed to shadowBan someone.

This is COMPLETELY AGAINST THE IDEA OF FREE SPEECH. A group of criminals or virtual militias can now SILENCE someone they don't like, and this ends up being perceived by the user because it is now possible to see the view statistics from someone's tweets and replies.

A lot of people are being unfairly silenced by malicious groups now, because that they have access to this code and now understands how it works and how to do it.

### DIFF
--- a/src/scala/com/twitter/interaction_graph/scio/agg_negative/InteractionGraphNegativeJob.scala
+++ b/src/scala/com/twitter/interaction_graph/scio/agg_negative/InteractionGraphNegativeJob.scala
@@ -46,6 +46,20 @@ object InteractionGraphNegativeJob extends ScioBeamJob[InteractionGraphNegativeO
     opts: InteractionGraphNegativeOption
   ): Unit = {
 
+    /*There is clearly a PROBLEM with this aproach: ShadowBan is supposed to ban abusers from the plataform, but with
+    * this code, it can be used by malicious groups and virtual militias to produce a coordinated ATTACK to silence
+    * someone they don't like. There are several reports of this on Twitter, people who defend human rights or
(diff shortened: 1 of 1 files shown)


ff137:
The following changes would make sense:
- unfollows are only counted when follow exists for more than x amount of days (I'm not sure how the `NumUnfollows` feature currently counts).
- mutes and blocks are only counted when it comes from accounts older than a certain number of days, or some other identifier to not count mutes/blocks from bot accounts

TwitterScrub:
I run an art account. I told several people I wouldn’t do adult artwork of characters under 18, fictional or not. They proceeded to attack my account directly and do exactly this to bury my account. I went from making 8k followers in several months to almost zero growth in a year.

From 10k views a post in an hour to 500views in five hours.

There appears to be no option to reset once this damage has been done, but we can at least try and make sure it can’t be done to someone else in the future.

amark:
I'm having similar issues, I have a small/moderate account (5.5K followers) but after getting blocked by Yud (AI doom guy) my views went from ~2K to 200.

This PR doesn't seem to address any code though, just a comment? Do people know of another PR or source code location of downranking blocked accounts?

PR #1801, opened 2023-04-21
Improve code readability and performance by using f-strings

Liam-hi:
- F-strings improve the readability 
- F-strings are often faster than concatenation because they are evaluated at runtime rather than being built up piece by piece.


ubergard:
LGTM

PR #1813, opened 2023-04-30
Remove unused variables

mzarchi:
There were two variables in the code that were used only once, out of respect for Twitter developers, I removed this variable and entered the values directly into dict.

### DIFF
--- a/twml/twml/contrib/build_graphs_fns.py
+++ b/twml/twml/contrib/build_graphs_fns.py
@@ -6,27 +6,28 @@ import tensorflow.compat.v1 as tf
 
 
 def get_saved_modules_graph(input_graph_fn):
-  """
-  Get common graph for stitching different saved modules for export.
-  This graph is used to save checkpoints; and then export the modules
-  as a unity.
-  Args:
-        features:
-          model features
-        params:
-          model params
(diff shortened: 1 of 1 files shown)


ubergard:
LGTM

gran4:
Next time, keep in mind it is usually better to keep the git history clean. Your commit shows that the whole section was deleted then re-added. I guess its fine for now. lgtm

mzarchi:
> Next time, keep in mind it is usually better to keep the git history clean. Your commit shows that the whole section was deleted then re-added. I guess its fine for now. lgtm

Ok, thanks for it ❤️

mzarchi:
> > Next time, keep in mind it is usually better to keep the git history clean. Your commit shows that the whole section was deleted then re-added. I guess its fine for now. lgtm
> 
> That is because he moved the whole code section on step further to the right, instead of just editing the values that needed to be edited. @mzarchi i recommend fixing your broken code indentation so the twitter devs have a proper change diff.

Be sure and thanks for the beauty point you said ❤️

mzarchi:
> LGTM

done ✅

RiteshShakya:
LGTM

PR #1828, opened 2023-05-18
Help protect Smirnoff!

TolstoyDotCom:
Smirnoff is a much-beloved and iconic American brand. However, lately a small group of vile trolls have been spreading disinfo and hate against the brand. I call on Twitter to stop this!

Smirnoff has done their best by hiding all of the trolls' hate-filled replies on threads like [this](https://twitter.com/SmirnoffUS/status/1653785214529806336). That works for logged-in users: they can't see any of the hidden replies. However, those who are logged out can have their minds polluted with some of the worst hate around.

Twitter must let cherished corporations like Smirnoff silence hate to not just logged-in users, but to those who are logged out. Twitter must go even further and ban any user who engages in anti-Smirnoff hate.

Corporations are America's most cherished institutions; they're our last line of defense against anarchy. They must be protected, allowed to control debate, and allowed to silence those spreading hate. Some will say that Twitter fully understands that and fully agrees and they're largely right. However, Twitter is not doing enough. Twitter must do more to stamp out hate!


Marioood:
LGTM

PatoFlamejanteTV (commented):
This is oddly specific, but it seems like a good idea to me.

Issue #1841, opened 2023-05-27
Warn Twitter Blue subscribers when they're being defrauded


TolstoyDotCom:
**Is your feature request related to a problem? Please describe.**
Twitter Blue subscribers are told their tweets will be at the top of lists. That's not the case for most of them. In fact, Twitter Blue subscribers are as heavily censored as other users.

**Describe the solution you'd like**
When a Twitter Blue subscriber tweets and the Twitter algorithm decides to censor that tweet, show a dialog box or similar letting the user know they were defrauded.

It didn't take long to find an example of how Twitter defrauds Twitter Blue subscribers. See the 'show more replies' section at /CAgovernor/status/1661839150646169601 ([screenshot](https://i.postimg.cc/nhFVnK8w/twitter-blue-censored.png)). Tweets in that section are censored, called "LowQuality" by Twitter. A Twitter Blue subscriber is in that section, below dozens of people that haven't paid Musk money.

I've filed a [complaint with the FTC](https://i.postimg.cc/tR820D6p/twitter-blue-ftc-complaint-obfuscated.png) about Twitter's fraud. I urge others to do the same, contact class action lawyers to see if they'd be interested in taking action, and also wising up Twitter Blue subscribers to how they're being defrauded.

Xpenzz:
Get a life you loser.

TolstoyDotCom:
Lucky us, we've had a visit from our local Musk fanboy. Let's see if we can help it try to make a valid argument for once.

What are the main reasons people sign up for Twitter Blue?

What does Twitter call the section after the 'show more replies' link?

falseuniversefacts:
Agreed

ErrorOliver:
I found a solution!

1. Go to 𝕏.
2. Go to account settings.
3. Delete account.
4. Create a account for Mastodon.
5. Use Mastodon.

Xpenzz:
> I found a solution!
> 
> 1. Go to 𝕏.
> 2. Go to account settings.
> 3. Delete account.
> 4. Create a account for Mastodon.
> 5. Use Mastodon.

Why would anyone sign up for that pedo platform https://www.washingtonpost.com/politics/2023/07/24/twitter-rival-mastodon-rife-with-child-abuse-material-study-finds/

Issue #54, opened 2023-03-31
why doesnt elons dad love him


enzo-js:
does anyone know? also why did he get his daughter pregnant?

enzo-js:
anyone know

airpods69:
man asking the real questions.

PR #1862, opened 2023-06-28
Correct improper promotion of low quality content

pandapawskat:
Turns out being able to pay to boost your content to the top of replies results in the people with the worst possible replies signing up. It's the only way they get seen because they're so low quality the algorithm naturally drops them out of view. So now that we have a simple way of identifying the lowest quality replies it's simple to change them to be forced to the bottom instead of the top. (I would have changed the scale parameters but I couldn't find them defined in this repo. But this should save looking up the parameters resulting in better algorithm performance.)

### DIFF
--- a/home-mixer/server/src/main/scala/com/twitter/home_mixer/functional_component/scorer/VerifiedAuthorScalingScorer.scala
+++ b/home-mixer/server/src/main/scala/com/twitter/home_mixer/functional_component/scorer/VerifiedAuthorScalingScorer.scala
@@ -3,8 +3,6 @@ package com.twitter.home_mixer.functional_component.scorer
 import com.twitter.home_mixer.model.HomeFeatures.AuthorIsBlueVerifiedFeature
 import com.twitter.home_mixer.model.HomeFeatures.InNetworkFeature
 import com.twitter.home_mixer.model.HomeFeatures.ScoreFeature
-import com.twitter.home_mixer.param.HomeGlobalParams.BlueVerifiedAuthorInNetworkMultiplierParam
-import com.twitter.home_mixer.param.HomeGlobalParams.BlueVerifiedAuthorOutOfNetworkMultiplierParam
 import com.twitter.product_mixer.component_library.model.candidate.TweetCandidate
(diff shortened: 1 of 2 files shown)


TolstoyDotCom:
Interesting POV. My [extensive research](https://tolstoy.com/twittercensorship/) shows that Twitter tends to promote junk while suppressing a good amount of higher quality content. Here are some [examples of how the Twitter algorithm does not work](https://tolstoydotcom.github.io/twitter-ranking-algorithm/).

Could you post something similar to the last showing at least anecdata to support your position? I mean examples of dreck that was promoted vs suppressed things further down that you think are higher quality.

Also, since [Twitter defrauds many blue subscribers](https://github.com/twitter/the-algorithm/issues/1841), wouldn't it be better for you to help wise up those who Musk is defrauding rather than having a little fit and trying to silence his victims?

uniformization:
So your solution is to rank Twitter Blue users under normal users? I would just remove the `isAuthorBlueVerified` check entirely.

TolstoyDotCom:
My solution has been readily available for years and I haven't been shy about promoting it. In less than an hour I came up with a tweet ranking formula that - IMNSHO - is far better than what Twitter does. The code is [here](https://github.com/TolstoyDotCom/more-speech) (see the userscripts directory). You can see it in action at the first link in my last comment. On each of the reports you'll see things like _original: #54, ranking: #213_. That means the Twitter algorithm - the one in this repo - put that tweet in 54th position but my algorithm doesn't think it's that good: it put it in 213rd place.

My algorithm is only based on the content of the tweet, not the easily manipulated things Twitter relies on (like blocks). Heck, even Musk has admitted blocking can be used as a form of DDS, not that he's doing anything about it.

uniformization:
> My solution has been readily available for years and I haven't been shy about promoting it…

Thanks for promoting your solution. Not to come off as rude in any way, but I was referring to @pandapawskat’s solution.

ALEX3955:
Yes

juicemission:
This is because the algorithm is not running arch btw

Issue #1386, opened 2023-04-02
Recommendation Algorithm Manipulation via mass blocks


redknightlois:
The current implementation allows for coordinated hurting of account reputation without recourse. The most general behavior is that global penalties are prone to be gamed (all of them). In other time I would just report this information using a vulnerability channel, but given that this is already popular knowledge there is no use to do so. 

The reason is that there is nothing a user can do to get rid of it because:
- The user can't know that it is been penalized.
- The user can't revert the penalty because it is not in his hands to change behavior to avoid it
- They accumulate and survive the actual tweet.
- No matter how much you boost, with enough people applying enough signals (there are many) the multiplier gets incredibly low.

**To Reproduce**
Organize  a botnet or a group of people with known similar views. 
Request your followers to block someone for 'reasons' (it doesn't matter here if the reasons are valid or not). This is exploited by political parties, group-think, etc. Now that this is also known, the vulnerability is plain obvious. 

Examples (using them to show the behavior does exist, not to punish the users for anything I had a lot to choose from):

https://twitter.com/BlockTheBlue
https://twitter.com/ayybeary/status/1642280442047995906
https://twitter.com/Kaptain_Kobold/status/1642379706925477888
https://twitter.com/MAYBEEELI/status/1642300879649792004
https://twitter.com/glenda_aus/status/1642282010462007296

There are apps that allow you build/organize/weaponize this behavior.
- https://www.blockpartyapp.com/
- https://www.reddit.com/r/GamerGhazi/comments/31s03s/twitter_blocklist_exchange/
- https://blocktogether.org/ 

While already shutdown, these are some of the stats for BlockTogether:
- 303k registered users.
- 198k users subscribing to at least one list.
- 4.5k users offering a list, with at least one subscriber.
- 3.7B actions.

1. Organize a group with a few friends (I have groups with 40+)
2. Find a target, and execute the following tasks in order
3. They should follow in preparation, a few days later unfollow first, [just doing this in 90 days intervals also hurts]
4. Then they will report a few "borderline" posts.
5. Then they will mute.
6. Then they will block. 

**Expected behavior**
No global penalty should be applied because you can game them pretty easily, all penalties (if any) should be applied at the content level.

jbauernberger:
Someone did a CVE already. What a time to be alive. 
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-29218

chaoscode:
So we are opening CVE's for cancel culture now?

[...]

PR #1882, opened 2023-07-29
InteractionGraphNegativeJob: accountLevelShadowbanMaxTime = 1 (in days)

bairesearch:
Limits all account level "shadow-banning" caused by blocks/mutes/unfollows to X days (configurable);

InteractionGraphNegativeJob.scala: `accountLevelShadowbanMaxTime = 1 //in days`

Rationale;
Account level censorship based on block/mute feature is easily gamified. Any censorship on this platform should be meticulously defined in the rules and performed at a content level.

See also;
https://github.com/twitter/the-algorithm/issues/1386
https://github.com/twitter/the-algorithm/issues/658 
https://github.com/twitter/the-algorithm/pull/660


redknightlois:
Wont solve the problem. But this is better than the 90 days solution.

amark:
As I spoke [here](https://github.com/twitter/the-algorithm/pull/660#issuecomment-1532429260), correct, any ability for blocks/etc. to affect the algorithm's distribution of content to *other* people is a messed up way that has allowed for bad actors to collude to oppress & suppress competing views. It shouldn't be coded at all.

Lisag123:
Is this going to be changed where blocks/mutes/reports have a time limit to prevent permanent shadowban?

Dakad:
ddsddfg

Issue #871, opened 2023-04-01
add my cat into the algorithm


TruncatedDinosour:
in response of #869 

![box_cat](https://user-images.githubusercontent.com/71613062/229289159-8dcb3b69-ed44-4c67-b276-ca2f350d4108.jpg)

her name is tina she would love to get added

hopingsteam:
Is already planned in #874

TruncatedDinosour:
> Is already planned in #874

thats a CI/CD supervisor, tina would like to be in charge of the source code, what gets merged and whatnot

Naim2000:
I dont wanna register on twitter does anyone else see the cat because this was closed as completed

TruncatedDinosour:
> I dont wanna register on twitter does anyone else see the cat because this was closed as completed

i still see the cat

![box_cat](https://user-images.githubusercontent.com/71613062/229310751-2a0cd268-3e73-47b5-8772-591c3d13278a.jpg)
![pov_my_cat_kisses_you](https://user-images.githubusercontent.com/71613062/229310767-ba906bc7-7eb9-4bb0-bb63-1094b9a7ef32.jpg)

heres some flattering pictures of tina though

TruncatedDinosour:
![6001086812195502382_121](https://user-images.githubusercontent.com/71613062/229311069-48872a3a-4c5d-4dd7-b1a0-434bd2fa1e1b.jpg)
![5890782826444470994_121](https://user-images.githubusercontent.com/71613062/229311073-16d8d46a-6789-4ee1-bf20-00052c53bef1.jpg)
![5857419078107512557_121](https://user-images.githubusercontent.com/71613062/229311074-39d368eb-87c5-4299-b897-5e3ad9d85d28.jpg)
![5857419078107512556_121](https://user-images.githubusercontent.com/71613062/229311077-70907a27-6d3f-4844-a2d5-1a69550e28d5.jpg)
![5766988258222391858_121](https://user-images.githubusercontent.com/71613062/229311078-2e7377da-4f3c-4953-aa0d-e3b41c1c566d.jpg)
![5766988258222391857_121](https://user-images.githubusercontent.com/71613062/229311079-046f2ace-3c88-41c9-a900-aa37079223c6.jpg)


heres a few more

TruncatedDinosour:
in conclusion, this should be reopened and discussed further bc i think this is
a really important step to ensure twitters security, maintenance, efficiency
[...]

PR #1941, opened 2023-12-18
fix https://github.com/twitter/the-algorithm/issues/871

ghost:
q


ar1ja (commented):
as the author of #871 i do agree that this issue is important and in need of addressing
lgtm

ghost:
added a fix for everything wrong with twitter

PR #1946, opened 2024-01-02
Update Reputation.scala

UnknownDeveloper240:
I increased num friends from 2500 to 9999 because why the hell not.

This fixes an important exploit that would've gave third-party attackers server-side access


PatoFlamejanteTV (commented):
That's actually a pretty good change, but I think it can mess and slow down the "AI Feed"

PR #1965, opened 2024-01-21
README.md looks better (1 line changed pls)

IyaadArshad:
Changed 1 line to appear more friendly in the README.md

Pls accept this change


PatoFlamejanteTV (commented):
I think it don't make sense, like I think the original/unedited version look better, idk it seems more professional for me.

PR #15986, opened 2025-09-12
Removed duplicated branching

nag2mani:
Removed duplicated branching with getDelegate(numPostings).

Optimizations: Avoid redundant branching;
Both copyPostingList and postings perform the same branching (numPostings < dfThreshold ? lowDF : highDF).

Extract that into a helper method getDelegate(numPostings).


nag2mani:
sorry, changed some other code as well.

PR #15988, opened 2025-09-13
docs: make it better to copy and past the path

1uc4sm4theus:


Snielsen49:
How exactly does this help with copy and paste?

PR #15990, opened 2025-09-13
Enhancement: Optimized embedding weights initialization to reduce memory allocations

pawannn:
# Description
This PR optimizes the creation of the weights array in embedding_lookup.py.

**Before**:
```python
weights = list(vocab.values())
weights = np.asarray(weights, dtype=np.float32)
```
- Built a full Python list and then converted it to a NumPy array, causing two memory allocations and unnecessary copying.

**After**:
```python
weights = np.empty((len(vocab), embedding_size), dtype=np.float32)
for idx, vec in enumerate(vocab.values()):
    weights[idx] = np.asarray(vec, dtype=np.float32)
```
- Preallocates the NumPy array and fills it in-place.
- Reduces memory usage and speeds up initialization for large vocabularies.

**Impact**:
- More memory-efficient embedding loading
- Faster initialization when working with large vocabularies
- No changes to external behavior or API


pawannn:
Hello Anyone there 🫤

PR #15995, opened 2025-09-19
Tweepcred: Clamp + Entropy Guard to prevent coordinated lists & brigades

FaxanFM:
### Changes

- Introduced maxDailyNegDelta, normalCap, and attackCap constants.
- Added an entropy/diversity guard: if fewer than 50% of issuers are unique, classify the event as a coordinated brigade.
- Implemented dynamic clamping: normal cap = –100, brigade cap = –20.
- Included telemetry stubs for brigade detection and clamp activation.

### Impact

- Prevents a small group of accounts from collapsing a user’s Tweepcred in a single batch run.
- Protects small and new creators from brigading while still applying legitimate, organic penalties.
- Preserves organic feedback fidelity (diverse, independent block signals continue to count).
- Provides clear telemetry for monitoring and future adjustments.

### Notes 

- Backward-compatible: retains the original one-line clamp.
- New constants are scoped private with safe defaults.
- This patch only affects Tweepcred batch reputation adjustment; no changes to visibilitylib or ranking layers.


kehlsteinhaus:
Did you really use ChatGPT to make this pull request?

PR #16101, opened 2026-09-07
Keep author and source ids when TES or cache misses

Pitchfork-and-Torch:
HasAuthorFilter drops a Home scored tweet when `AuthorIdFeature` is empty. Tweet Entity Service (TES) and the static-entity cache used an empty default map on a miss, so a post that already had an author from Earlybird / TimelineRanker or from tweet-mixer was treated as having no author and removed.

This is not GrokFilter (#4), not the tweet-mixer Tweetypie cache-miss path (#2), and not a visibility-label expiry/country change (#1, #3).

## Evidence

`TweetEntityServiceFeatureHydrator` on cache or TES miss (including a PureCoreData miss, which drops the whole TES record):

- wrote `DefaultFeatureMap` with `AuthorIdFeature = None`
- also cleared `SourceTweetIdFeature` / `SourceUserIdFeature` / reply and quote ids

`HasAuthorFilter` then removes the candidate. Downstream scoring uses those source ids as the original entity, so a wipe also retargets the post.

`ScoredTweetsTweetMixerResponseFeatureTransformer` never copied `TweetResult.authorId`. tweet-mixer already resolved the post author. `tweetMetadata.authorId` is the signal author (the user whose like or follow sourced the candidate). Those are different ids. After the transformer drop, TES was the only remaining author source for out-of-network tweet-mixer candidates.

The rest of the same path already keeps known ids on a miss:

- `TweetypieStaticEntitiesFeatureHydrator` keeps `AuthorIdFeature` from the candidate source when Tweetypie does not return Found

## Change

- On TES or cache miss, keep author, source, reply, quote, and retweet ids the candidate source already set. Empty only when the source had none.
- Copy `TweetResult.authorId` onto `AuthorIdFeature`. Do not use `tweetMetadata.authorId` for that field.

TES hits still overwrite from the store. No write-path, TTL, or flag changes.

## Verify

```
python3 home-mixer/src/test/tes_store_miss_author_id_test.py
```

home-mixer has no local Scala harness for this hydrator (Strato / Tweetypie clients). The Python file is the decision table for miss-keep vs HasAuthorFilter drop.

| TES / cache | Existing author | After | HasAuthorFilter |
|---|---|---|---|
| miss | 42 | 42 | keep |
| miss | none | none | drop |
| hit | 42 | store author | keep if store has author |

## Steward / x-algorithm

Same class likely exists in `xai-org/x-algorithm`. Open PR #94 there is TES core-data miss clearing retweet and reply ids. Steward should port this twin as well: a TES miss must not wipe author id into a drop filter, and the For You candidate ingest must keep the post author, not the signal author. Confirm #94 covers author-id wipe, not only retweet/reply ids.

<div><a href="https://cursor.com/agents/bc-e06d7e02-4e4b-4731-9720-23ef9b40ffdc?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/ima

Fork mirror closed; this is the ship vehicle.


Pitchfork-and-Torch:
Wrong repo. Serving-path hunt is xai-org/x-algorithm (2026 dump), not twitter/the-algorithm (2023 Java dump). Closing. Do not reopen.

Pitchfork-and-Torch:
Wrong repo. Serving-path hunt is xai-org/x-algorithm (2026 dump), not twitter/the-algorithm (2023 Java dump). Closing. Do not reopen.

